PR #19, opened 2018-01-02
Radiance tb conversions

### LINKED ISSUE #18: Enhanced feature: Allow to derive the band integrated radiance

```
from pyspectral.radiance_tb_conversion import RadTbConverter
from pyspectral.utils import debug_on
debug_on()
modis = RadTbConverter('EOS-Aqua', 'modis', '20')
res = modis.tb2radiance(281., '20', lut=False)

```
#### Problem description

When the above code is run, the spectral radiance for the MODIS band 20 is derived.
One issue is that it should actually not be needed to pass the band to the method. It should
actually not be allowed to derive the radiance for another band.

But, my main issue, is that it should also be possible to derive the band integrated radiance.
This can be derived by not dividing with the integral of the relative spectral response for the given 
band. So something like this:
```
modis = RadTbConverter('EOS-Aqua', 'modis', '20')
res = modis.tb2radiance(281., lut=False, normalized=False)

adybbroe:
Clean up in the code particularly concerning radiance to Tb conversions.
Allow to derive the band integrated radiance from a Tb, in addition to the default spectral radiance.
Separate the general case using RSR data from the specific alternative case for SEVIRI (so far only)
where the radiance to Tb derivations can also be done using off line coefficients from EUMETSAT

 - [x] Closes #18
 - [x] Tests added
 - [x] Tests passed
 - [x] Passes ``git diff origin/develop **/*py | flake8 --diff`` 
 - [x] Fully documented


adybbroe:
Done.

PR #53, opened 2018-11-01
Fix masking of calculated NIR reflectances

pnuu:
This PR fixes masking of the calculated NIR reflectances.

 - [ ] Tests passed 


pnuu:
Already forgot why the zeros need to be removed, so ran few scenes again to see the effect, and here's one comparison:

Image with `master` branch (zeros left in):
image

Same section with this PR in use:
image

So the bad data inside the swath is replaced with zeros, which is an unwanted side-effect. With normal composites not using NIR reflectance, e.g. `overview`, both branches yield a result that has the gap masked out.

mraspaud:
Nice example ! However, I'd like to argue that the bad values should already be flagged and marked as NaN in the reader. At the NIR reflectance correction stage, we should assume that the non-NaN values are valid.

pnuu:
The values _are_ marked as NaN by the reader, as happens with the non-pyspectral SatPy composites. So the zero-filling happens when pyspectral NIR reflectance is calculated:

image

pnuu:
Yes. Must be the casting from dask arrays to masked arrays and other way around that causes this. This PR fixes the result at least. Pyspectral has a zenith angle limitation, so the zero reflectance can't be reached (no data calculated for the night side). The next step would be to use pure dask arrays all the way, but don't want to start working on that as this small fix works for now.

pnuu:
Changed the masking as suggest by @mraspaud in Slack. Also updated the topic to better reflect what is happening.

adybbroe:
Thanks @pnuu Looks good.

PR #60, opened 2018-11-30
Fix readthedocs

adybbroe:
Fixing the ReadTheDoc pages. RTD doesn't support git-lfs, and we added some png's recently in git-lfs.

 - [x] Tests passed 

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ except IOError:
 requires = ['docutils>=0.3', 'numpy>=1.5.1', 'scipy>=0.14',
             'python-geotiepoints>=1.1.1',
             'h5py>=2.5', 'requests', 'six', 'pyyaml',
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -37,7 +37,7 @@
 requires = ['docutils>=0.3', 'numpy>=1.5.1', 'scipy>=0.14',
             'python-geotiepoints>=1.1.1',
             'h5py>=2.5', 'requests', 'six', 'pyyaml',
-            'appdirs']
+            'appdirs', 'git-lfs']
```
djhoese: If the git LFS is only needed for the docs then shouldn't this maybe be moved an extra_requires or be conditional on being on the RTD environment?

adybbroe: Agree, fixed. Had forgot my rtd_requirerments.txt file

PR #64, opened 2019-02-11
Fix interp function in rayleigh correction to be serializable

djhoese:
When using dask in a distributed way, functions passed to `map_blocks` and `delayed` can't be internal functions (functions defined inside other functions) because they are not serializable/accessible in a way that dask/multiprocessing can use. This PR refactors the rayleigh correction to not use an internal function. I'm not really sure how to write a test for this.

 - [ ] Tests added 
 - [x] Tests passed 


djhoese:
@adybbroe Yes. I was generating true color images on a cluster. It only moves the function so I'd be surprised if this breaks anything else.

PR #74, opened 2019-05-03
Fix dask compatibility in blackbody functions

### LINKED ISSUE #73: Fix blackbody code to work with dask arrays
There are a lot of functions in `pyspectral.blackbody` that call `np.array(some_input_argument)` which can unnecessarily convert dask arrays to numpy arrays. I/we should update this so it allows for dask arrays to pass through. One option:

```
try:
    from dask.array import asanyarray
except ImportError:
    from numpy import asanyarray
```

Thoughts?

djhoese:
I was writing the MERSI-2 L1B reader and needed to use the rad2temp function. I noticed that when dask arrays were passed they were being converted to a numpy array. This PR fixes the usage with the least amount of changes possible.

The biggest difference is that in the `planck` function a masked array was being used with `-9` as the original sentinel value. As far as I could tell the masked array was being used for two reasons: to make `min/max` calls ignore the masked values and to later use the `.mask` to replace masked values with NaN. I switched this to use NaN right away instead of `-9` and to only print out min/max debug information when the input is an numpy array.

I'm marking this as a bug fix because although its a feature, it is more about fixing a working function so it returns what I want.

 - [x] Closes #73 
 - [x] Tests added 
 - [x] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/pyspectral/blackbody.py
+++ b/pyspectral/blackbody.py
@@ -23,8 +23,13 @@
 
 """Planck radiation equation"""
 import numpy as np
-
 import logging
+
+try:
+    import dask.array as da
+except ImportError:
+    da = np
+
 LOG = logging.getLogger(__name__)
 
 H_PLANCK = 6.62606957 * 1e-34  # SI-unit = [J*s]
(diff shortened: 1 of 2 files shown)


adybbroe:
@pnuu Can you have a look as well before I merge?

PR #93, opened 2019-11-29
Abandon gitlfs

### LINKED ISSUE #92: Fails to pull from Github because of a git-lfs quota limit

```python
# Your code here

```
#### Problem description

Travis just started failing this week without any changes being done to the repo. It appears we have reached a limit on the git-lfs quota. We need to find a way to handle this!

See [here](https://github.com/nabla-c0d3/nassl/issues/17)

adybbroe:
- [x] Closes #92  
 - [ ] Tests added 
 - [x] Tests passed 
 - [ ] Passes ``git diff origin/master **/*py | flake8 --diff`` 
 - [ ] Fully documented


djhoese:
Could/should these PNGs be removed from the entire git history so they don't take up any space?

PR #95, opened 2019-12-06
Adapt for later versions of TQDM

adybbroe:
Adjust ftp download chunk size and the number of expected iterations to an integer.

In the most recent version of tqdm (4.40 as opposed to 4.37) pyspectral fails if the total number of iterations when downloading data in chunks is not an integer. 

tqdm is used to show status bar when downloading LUTs via ftp.

Thanks to @aronnem for reporting this!

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented

### DIFF
--- a/pyspectral/utils.py
+++ b/pyspectral/utils.py
@@ -433,7 +433,7 @@ def download_luts(**kwargs):
     else:
         aerosol_types = HTTPS_RAYLEIGH_LUTS.keys()
 
-    chunk_size = 4096
+    chunk_size = 1024*1024  # 1 MB
 
     for subname in aerosol_types:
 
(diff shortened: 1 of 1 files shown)


pyspectral/utils.py
```diff
@@ -433,7 +433,7 @@ def download_luts(**kwargs):
     else:
         aerosol_types = HTTPS_RAYLEIGH_LUTS.keys()
 
-    chunk_size = 10124
+    chunk_size = 4096
```
djhoese: According to https://requests.readthedocs.io/en/master/api/#requests.Response.iter_content the chunk size is number of bytes so this is only 4K. I think we want something in the range of 1-10MB.

adybbroe: I just fixed that, you are too quick @djhoese ! :-)

adybbroe: We are downloading files in the order of 3-75 MB

PR #109, opened 2020-05-11
add more realistic METimage RSRs

rolle-at-work:
Updated to more realistic, though not measured spectral response functions for METimage

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented

### DIFF
--- a/pyspectral/utils.py
+++ b/pyspectral/utils.py
@@ -175,11 +175,11 @@ INSTRUMENTS = {'NOAA-19': 'avhrr/3',
                }
 
 
-HTTP_PYSPECTRAL_RSR = "https://zenodo.org/record/3813522/files/pyspectral_rsr_data.tgz"
+HTTP_PYSPECTRAL_RSR = "https://zenodo.org/record/3824535/files/pyspectral_rsr_data.tgz"
 
 
 RSR_DATA_VERSION_FILENAME = "PYSPECTRAL_RSR_VERSION"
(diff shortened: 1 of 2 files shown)


rolle-at-work:
The new file looks good and i tested reading with hdfview and a pyspectral plotting routine.

rolle-at-work:
@adybbroe But is this layer also gone for AVHRR and other one-detector instruments? I think this layer was usefull to allow for a general reader of bof both single and multi detector instruments.

adybbroe:
Yes @rolle-at-work they are gone for all single detector sensors. They should be at least. I just checked a couple. And the reader can handle both, no problem

adybbroe:
Oh well, now I look back in the history here: 
https://github.com/pytroll/pyspectral/pull/68
Seems like we were preparing for more detectors. So, if we know there will be, and I gues we do know that @rolle-at-work? Then we should keep it as is I think. Sorry

rolle-at-work:
Yes @adybbroe METimage will have 24 detectors per channel

adybbroe:
@rolle-at-work I have uploaded an updated RSR data archive with your new Metimage data included, please check it out and adapt your PR accordingly. You will need to change version and url of the RSR data in `utils.py` Version should now be `RSR_DATA_VERSION = "v1.0.15"` and the url is now https://zenodo.org/record/3824535/files/pyspectral_rsr_data.tgz

PR #113, opened 2020-10-05
Separate masking and Sun zenith angle correction

### LINKED ISSUE #112: Do not cut NIR reflectance with a single threshold

```python
import glob
from satpy import Scene

fnames = glob.glob("/path/to/msg/data/H*__")
glbl = glbl.load(["day_microphysics"])
lcl = glbl.resample("euro4")
lcl.show("day_microphysics")
```

#### Problem description

The above code using Satpy uses Pyspectral to determine the NIR reflectance and uses it as one channel in the RGB. Current version of Pyspectral masks the data with a single strict limit of Sun zenith angle (85.0 degrees by default). The same limit is used to limit the range of SZA correction. These two should be separated. When this composite is used together with a night-side composite, there's a big gap between the composites.

#### Expected Output
Adjustable limit for masking invalid data by SZA.

day_microphysical_20161128_110000

#### Proposed solution
Separate the SZA limit (`sunz_threshold` kwarg) and the masking limit of invalid data. So I would add another keyword argument to `pyspectral.near_infrared_reflectance.Calculator.__init__`. Maybe `masking_limit`. By default these both can have the same limit (`TERMINATOR_LIMIT`), but for Day/Night composites the masking limit should be adaptable.

Current master branch of Pyspectral.

pnuu:
This PR separates the Sun zenith-angle correction and masking of data beyond given SZA limit.

 - [x] Closes #112  
 - [x] Tests added 
 - [x] Fully documented

### DIFF
--- a/doc/seviri_example.rst
+++ b/doc/seviri_example.rst
@@ -27,7 +27,7 @@ reflectance, saving a few milliseconds per call::
 
 By default the data are masked outside the default Sun zenith-angle (SZA) correction limit (85.0 degrees).
 The masking can be adjusted via `masking_limit` keyword argument to `Calculator`, and turned of by
-defining `Calculator(..., masking_limit=None)`.  The SZA limit can be adjusted via `sunz_threshold keyword argument:
+defining `Calculator(..., masking_limit=None)`.  The SZA limit can be adjusted via `sunz_threshold` keyword argument:
 `Calculator(..., sunz_threshold=88.0)`.
 
 Integration with SatPy
--- a/doc/seviri_example.rst
+++ b/doc/seviri_example.rst
@@ -25,6 +25,10 @@ reflectance, saving a few milliseconds per call::
   >>> print('%4.3f' %refl39.reflectance_from_tbs(sunz, tb3, tb4))
   0.555
 
+By default the data are masked outside the default Sun zenith-angle (SZA) correction limit (85.0 degrees).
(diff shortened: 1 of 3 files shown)


pnuu:
And here's an example using a matching Satpy PR (yet-to-be-PR'd) branch:

day_microphysical_with_night_microphysical_20200320_180000

This provides much better looking results than the ones in #112

pnuu:
Travis tests fail here: https://github.com/pytroll/pyspectral/blob/master/pyspectral/tests/test_reflectance.py#L181
The same test passes locally, don't know what's going on..

pnuu:
I had `numpy` 0.17.5 so tried upgrading to `0.19.1` which Travis used: still passes locally.

mraspaud:
@pnuu the error seems to come from conda failing on python 2.7, so I don't there is much you can do about it.

pnuu:
Hmm. I'm quite sure the other three jobs had a failure in the test I linked above, but now they've passed. Weird. Didn't notice the erroring test is from Python 2.7, makes sense it errors as it's not supported by Travis anymore.

mraspaud:
Just opened #114 to remove 2.7 testing.

adybbroe:
Sorry about your troubles @pnuu . @mraspaud is right, python 2.7 testing on Travis has failed a long time now. #116 should solve this. We leave Python 2.7 alltogether

adybbroe (approved):
LGTM

adybbroe:
Thanks for the nice work @pnuu !

PR #120, opened 2020-10-18
Fix scale of AGRI response values

simonrp84:
Within pyspectral, the response functions for instruments are in the range 0->1.
However, the raw AGRI spectral responses are scaled between 0->100 and must be converted into the correct range during processing. This PR updates the AGRI RSR reader to perform this conversion.


adybbroe (approved):
LGTM

pnuu (approved):
LGTM

PR #128, opened 2021-02-02
Fix dask compatibility of IR atmospherical correction

ch-k:
The IR atmospherical correction algorithm used by DWD is not compatible with dask arrays. Since dask is used already for some time by satpy, it has to be fixed. 
The provided fix is already used by the German weather service (DWD). Would be great if this could be merged to the upstream repo.

 - [x] Tests ~~added~~ modified 

### DIFF
--- a/pyspectral/atm_correction_ir.py
+++ b/pyspectral/atm_correction_ir.py
@@ -159,5 +159,5 @@ if __name__ == "__main__":
     NDIM = SHAPE[0] * SHAPE[1]
     SATZ = np.ma.arange(NDIM).reshape(SHAPE) * 60. / float(NDIM)
     TBS = np.ma.arange(NDIM).reshape(SHAPE) * 80.0 / float(NDIM) + 220.
-    atm_corr = this.get_correction(SATZ, 'M4', TBS)
+    atm_corr = this.get_correction(da.from_array(SATZ), 'M4', da.from_array(TBS))
 
--- a/pyspectral/atm_correction_ir.py
+++ b/pyspectral/atm_correction_ir.py
@@ -59,6 +59,7 @@ from this time are not longer at DWD.'
 
 
 import numpy as np
+import dask.array as da
 import logging
 
 LOG = logging.getLogger(__name__)
@@ -127,15 +128,38 @@ def viewzen_corr(data, view_zen):
         Z_REF = 70.0
         DELTA_REF = 6.2
         return (1 + DELTA_REF)**ratio(z, Z_0, Z_REF) - 1
-
-    y0, x0 = np.ma.where(view_zen == 0)
-    data[y0, x0] += tau0(data[y0, x0])
-
-    y, x = np.ma.where((view_zen > 0) & (view_zen < 90) & (~data.mask))
-    data[y, x] += tau(data[y, x]) * delta(view_zen[y, x])
-    return data
-
-
+      
+    # not allowed for dask arrays:
+    # - assignment
+    # - access with indexes
+    #
+    # that's why all calculations are done as numpy arrays
+    # special data conversion are needed    
+    # conversion from dask arrays into numpy arrays
+    LOG.debug("Update to process dask array data.") 
+    np_view_zen = np.array(view_zen.compute())
+    np_data = np.ma.masked_array(data.compute())
+    
+    #y0, x0 = np.ma.where(view_zen == 0)
+    y0, x0 = np.ma.where(np_view_zen == 0)
+        
+    # data[y0, x0] += tau0(data[y0, x0])
+    np_data[y0, x0] += tau0(np_data[y0, x0])
+    
+    # y, x = np.ma.where((view_zen > 0) & (view_zen < 90) & (~data.mask))
+    y, x = np.ma.where((np_view_zen > 0) & (np_view_zen < 90) & (~np_data.mask))
+    
+    # data[y, x] += tau(data[y, x]) * delta(view_zen[y, x])
+    np_data[y, x] += tau(np_data[y, x]) * delta(np_view_zen[y, x])
+            
+    LOG.debug( 'Calculated data min/avr/max: {:.3f}/{:.3f}/{:.3f}'.format(np.nanmin(np_data),  \
+                                                                          np.nanmean(np_data), \
+                                                                          np.nanmax(np_data)))
+                                              
+    # back conversion from numpy array into dask array
+    # return data                                              
+    return da.from_array(np_data)
+    
 if __name__ == "__main__":
     this = AtmosphericalCorrection('Suomi-NPP', 'viirs')
     SHAPE = (1000, 3000)
@@ -143,3 +167,4 @@ if __name__ == "__main__":
     SATZ = np.ma.arange(NDIM).reshape(SHAPE) * 60. / float(NDIM)
     TBS = np.ma.arange(NDIM).reshape(SHAPE) * 80.0 / float(NDIM) + 220.
     atm_corr = this.get_correction(SATZ, 'M4', TBS)
+
--- a/pyspectral/atm_correction_ir.py
+++ b/pyspectral/atm_correction_ir.py
@@ -128,15 +128,15 @@ def viewzen_corr(data, view_zen):
         Z_REF = 70.0
         DELTA_REF = 6.2
         return (1 + DELTA_REF)**ratio(z, Z_0, Z_REF) - 1
-      
+
     # not allowed for dask arrays:
     # - assignment
     # - access with indexes
     #
     # that's why all calculations are done as numpy arrays
-    # special data conversion are needed    
+    # special data conversion are needed
     # conversion from dask arrays into numpy arrays
-    LOG.debug("Update to process dask array data.") 
+    LOG.debug("Update to process dask array data.")
     np_view_zen = np.array(view_zen.compute())
     np_data = np.ma.masked_array(data.compute())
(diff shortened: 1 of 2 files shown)


djhoese (changes requested):
Thanks for contributing back changes used at DWD. From what I can tell this isn't backwards compatible. You make it so dask is required, but this shouldn't be necessary/required if someone is using numpy. Instead of creating a new test to test dask array support, you've modified the existing test. Could you please add a new test so there is the old test for numpy support and a new test for dask support?

Lastly, you use `.compute()` in the function and then convert the result back to a dask array. This is very bad for overall performance and is more of a workaround for dask use rather than a complete "adoption" of its functionality. I am not @adybbroe so he may have other suggestions, but it be possible for you to either: create a separate function that gets called for dask arrays or (my preference) rewrite this function to use functions like `np.where` (or maybe `da.where`) to replace the existing masked array usage? Something like this:

```
zen_mask = view_zen == 0
data_tau = data + tau0(data)
data = da.where(zen_mask, data_tau, data)
```

For example. Or maybe instead of a new function you could do `if isinstance(data, np.ndarray)` to separate out dask versus numpy array usage.

ch-k:
> 
> 
> Thanks for contributing back changes used at DWD. From what I can tell this isn't backwards compatible. You make it so dask is required, but this shouldn't be necessary/required if someone is using numpy. Instead of creating a new test to test dask array support, you've modified the existing test. Could you please add a new test so there is the old test for numpy support and a new test for dask support?
> 
> Lastly, you use `.compute()` in the function and then convert the result back to a dask array. This is very bad for overall performance and is more of a workaround for dask use rather than a complete "adoption" of its functionality. I am not @adybbroe so he may have other suggestions, but it be possible for you to either: create a separate function that gets called for dask arrays or (my preference) rewrite this function to use functions like `np.where` (or maybe `da.where`) to replace the existing masked array usage? Something like this:
> 
> ```
> zen_mask = view_zen == 0
> data_tau = data + tau0(data)
> data = da.where(zen_mask, data_tau, data)
> ```
> 
> For example. Or maybe instead of a new function you could do `if isinstance(data, np.ndarray)` to separate out dask versus numpy array usage.

Yes, you are right. It isn't backwards compatible. I was not aware of this requirement. In my experience there is no compability to numpy is already broken in at least satpy. But of course there might be lot of people using pyspectral in other use cases. 
Would you mind if I introduce a switch to support numpy and dask in parallel at first (in this PR)? Obviously nobody uses this functionallity with dask besides DWD.

djhoese:
I won't speak for @adybbroe, but Satpy is a separate project where it is the creator and user of the data. The pyspectral, pyresample, and trollimage libraries that Satpy uses still have their own users that may be passing numpy arrays (I know pyresample does for sure). Although pyspectral and trollimage may have fewer users outside of Satpy, I'm not sure we've made a decision to completely drop support for pure numpy arrays in these packages. Even if we did, it would be to completely adopt xarray DataArrays (with dask underneath) in a way that would not require computing the data as you've done in this PR. I think the main issue with pyspectral is that it uses a lot of numpy masked arrays which aren't really something dask does right now so there isn't a 1:1 transition. In most cases dask arrays can be passed to numpy functions just fine.

I'd be ok with you making this PR work for both numpy and dask arrays. If you can make it so the dask arrays don't have to be computed that would be great.

pyspectral/atm_correction_ir.py
```diff
@@ -128,12 +129,29 @@ def delta(z):
...
+    y, x = np.ma.where((np_view_zen > 0) & (np_view_zen < 90) & (~np_data.mask))
+    np_data[y, x] += tau(np_data[y, x]) * delta(np_view_zen[y, x])
+
+    LOG.debug('Calculated data min/avr/max: {:.3f}/{:.3f}/{:.3f}'.format(np.nanmin(np_data),
+                                                                         np.nanmean(np_data),
+                                                                         np.nanmax(np_data)))
+
+    # back conversion from numpy array into dask array
+    # return data
+    return da.from_array(np_data)
```
djhoese, line 154: Does this actually work? Don't you have to specify the chunk size?

pyspectral/atm_correction_ir.py
```diff
@@ -128,12 +129,29 @@ def delta(z):
...
+    else:
+        y0, x0 = np.ma.where(view_zen == 0)
+        data[y0, x0] += tau0(data[y0, x0])
+
+        y, x = np.ma.where((view_zen > 0) & (view_zen < 90) & (~data.mask))
+        data[y, x] += tau(data[y, x]) * delta(view_zen[y, x])
+
+        LOG.debug('Calculated data min/avr/max: {:.3f}/{:.3f}/{:.3f}'.format(np.nanmin(data),
+                                                                             np.nanmean(data),
+                                                                             np.nanmax(data)))
```
djhoese, line 153: Was this log message here before?

pyspectral/atm_correction_ir.py
```diff
@@ -128,12 +129,29 @@ def delta(z):
...
-    data[y0, x0] += tau0(data[y0, x0])
+    if hasattr(data, 'compute') or hasattr(view_zen, 'compute'):
+        is_dask_data = True
+    else:
+        is_dask_data = False
 
-    y, x = np.ma.where((view_zen > 0) & (view_zen < 90) & (~data.mask))
-    data[y, x] += tau(data[y, x]) * delta(view_zen[y, x])
-    return data
+    if is_dask_data:
```
djhoese, line 137: Code factor (the bot) makes some good points. You could replace the above `is_dask_data` with a single line (ex. `is_data_data = hasattr(data, 'compute') ...`.

pyspectral/atm_correction_ir.py
```diff
@@ -59,6 +59,7 @@
 
 
 import numpy as np
+import dask.array as da
```
djhoese, line 62: Could you maybe put this in a try/except block. If ImportError then `da = None`. You'd then need to update your `is_dask_data` boolean too.

pyspectral/atm_correction_ir.py
```diff
@@ -128,12 +129,29 @@ def delta(z):
...
 
-    y, x = np.ma.where((view_zen > 0) & (view_zen < 90) & (~data.mask))
-    data[y, x] += tau(data[y, x]) * delta(view_zen[y, x])
-    return data
+    if is_dask_data:
+        dask_data = da.where(view_zen == 0,
+                             data + tau0(data),
+                             da.where((view_zen > 0) & (view_zen < 90),
+                                      data + (tau(data) * delta(view_zen)),
+                                      data))
```
djhoese, line 142: Could you split this into multiple lines. It is kind of complex for a single call and a little hard to read.

djhoese (changes requested):
Thanks for putting the work to make this work for dask and numpy. Just a few style things and small concerns, otherwise I think this is good for backwards compatibility.

I'd still like @adybbroe to have the final say.

pyspectral/atm_correction_ir.py
ch-k: No, it wasn't. It's for debugging only (that's why it's has debug level :-) ) and only for numpy to avoid unnecessary computations for dask data. I can remove it if you want.

pyspectral/atm_correction_ir.py
ch-k: No problem, I will change it.

pyspectral/atm_correction_ir.py
djhoese: Yeah, removing it would be good. It requires numpy to iterate over the data three times just for a debug message which is not good.

pyspectral/atm_correction_ir.py
ch-k: done

pyspectral/atm_correction_ir.py
ch-k: done

pyspectral/atm_correction_ir.py
```diff
@@ -59,6 +59,14 @@
 
 
 import numpy as np
+try:
+    import dask.array as da
+    HAVE_DASK = True
+except ImportError:
+    da = None
+    HAVE_DASK = False
```
djhoese, line 67: I just realized, you don't need the `HAVE_DASK` as you could just check `da is None`, **BUT** you don't actually need that check later on anyway. If the data has a `compute` method then we know `da` must not be None because dask must have been used to create the dask arrays.

adybbroe: Agree

adybbroe:
@ch-k Many thanks for bringing this back to Pytroll/Pyspectral.

I don't this part of Pyspectral is used outside DWD - it was added at a PCW a few years back, and just briefly tested then. Anyway I see this part as solely to support Satpy. But I welcome the backward compatibility anyway!

pyspectral/atm_correction_ir.py
adybbroe: So, @ch-k just remove `HAVE_DASK` as suggested by @djhoese

ch-k: done

adybbroe:
Great thanks @ch-k 
Maybe just fix the last CodeFactor issue and I am fine with this. Then we can wait for @djhoese waking up to say his last bit

adybbroe (commented):
As said, when the remaining CodeFactor issue is fixed this looks good to me
Thanks for the PR!

djhoese (approved):
Thanks!

PR #129, opened 2021-04-06
Fix so reflectances can be derived from scalar inputs

adybbroe:
Fix so reflectances can be derived from scalar inputs (including sun zenith)

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented

### DIFF
--- a/pyspectral/near_infrared_reflectance.py
+++ b/pyspectral/near_infrared_reflectance.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 #
-# Copyright (c) 2014-2020 Pytroll developers
+# Copyright (c) 2014-2021 Pytroll developers
 #
 # Author(s):
 #
@@ -219,15 +219,9 @@ class Calculator(RadTbConverter):
         else:
             is_masked = False
 
-        if np.isscalar(tb_near_ir):
-            tb_nir = array([tb_near_ir, ])
-        else:
-            tb_nir = asanyarray(tb_near_ir)
-
-        if np.isscalar(tb_thermal):
-            tb_therm = array([tb_thermal, ])
-        else:
-            tb_therm = asanyarray(tb_thermal)
+        tb_nir = get_as_array(tb_near_ir)
+        tb_therm = get_as_array(tb_thermal)
+        sun_zenith = get_as_array(sun_zenith)
 
         if tb_therm.shape != tb_nir.shape:
             errmsg = 'Dimensions do not match! {0} and {1}'.format(
@@ -266,6 +260,7 @@ class Calculator(RadTbConverter):
             LOG.info('l_nir = %s', str(l_nir))
 
         LOG.debug("Apply sun-zenith angle clipping between 0 and %5.2f", self.masking_limit)
+
         sunz = sun_zenith.clip(0, self.sunz_threshold)
         mu0 = np.cos(np.deg2rad(sunz))
 
@@ -303,3 +298,14 @@ class Calculator(RadTbConverter):
         if is_masked:
             res = np.ma.masked_invalid(res)
         return res
+
+
+def get_as_array(parameter):
+    """Return parameter as a Dask or Numpy array. 
+
+    Parameter may be a scalar, a list or a Numpy/Dask array.
+    """
+    if np.isscalar(parameter):
+        return array([parameter, ])
+    else:
+        return asanyarray(parameter)
--- a/pyspectral/near_infrared_reflectance.py
+++ b/pyspectral/near_infrared_reflectance.py
@@ -300,12 +300,12 @@ class Calculator(RadTbConverter):
         return res
 
 
-def get_as_array(parameter):
-    """Return parameter as a Dask or Numpy array. 
+def get_as_array(variable):
+    """Return variable as a Dask or Numpy array. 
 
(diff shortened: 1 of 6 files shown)


adybbroe:
@pnuu or @djhoese Any view on this!? Can we/I merge it?

pyspectral/tests/test_reflectance.py
```diff
@@ -221,3 +215,33 @@ def test_reflectance(self):
...
+
+
+def test_get_as_array_from_scalar_input():
+    """Test the function to return an array when input is a scalar"""
+    res = get_as_array(1)
+
+    if hasattr(res, 'compute'):
+        assert res.compute()[0] == 1
+    else:
+        assert res[0] == 1
```
pnuu, line 227: Should this Numpy case be tested in every case? It should be possible by patching, something like this:
```python
with mock.patch('from pyspectral.near_infrared_reflectance.array', new=np.ndarray):
    res = get_as_array(1)
    assert res[0] == 1
```
It might be necessary to move the imports inside the testing functions. Which is better practice anyway when patching is needed.

pyspectral/tests/test_reflectance.py
```diff
@@ -221,3 +215,33 @@ def test_reflectance(self):
...
+
+
+def test_get_as_array_from_numpy_array_input():
+    """Test the function to return an array when input is a numpy array."""
+    res = get_as_array(np.array([1.0, 2.0]))
+
+    if hasattr(res, 'compute'):
+        np.testing.assert_allclose(res.compute(), np.array([1.0, 2.0]), 5)
+    else:
+        np.testing.assert_allclose(res, np.array([1.0, 2.0]), 5)
```
pnuu, line 237: Same here. This pure-Numpy functionality should be tested every time.

pyspectral/tests/test_reflectance.py
```diff
@@ -221,3 +215,33 @@ def test_reflectance(self):
...
+
+
+def test_get_as_array_from_list_input():
+    """Test the function to return an array when input is a list."""
+    res = get_as_array([1.0, 2.0])
+
+    if hasattr(res, 'compute'):
+        np.testing.assert_allclose(res.compute(), np.array([1.0, 2.0]), 5)
+    else:
+        np.testing.assert_allclose(res, np.array([1.0, 2.0]), 5)
```
pnuu, line 247: And again, this pure-Numpy functionality should be tested every time.

pnuu (commented):
LGTM, just one (repeated) suggestion on testing also the Numpy implementation even if `dask` is available.

pyspectral/near_infrared_reflectance.py
```diff
@@ -219,15 +219,9 @@ def reflectance_from_tbs(self, sun_zenith, tb_near_ir, tb_thermal, **kwargs):
...
-        else:
-            tb_nir = asanyarray(tb_near_ir)
-
-        if np.isscalar(tb_thermal):
-            tb_therm = array([tb_thermal, ])
-        else:
-            tb_therm = asanyarray(tb_thermal)
+        tb_nir = get_as_array(tb_near_ir)
+        tb_therm = get_as_array(tb_thermal)
+        sun_zenith = get_as_array(sun_zenith)
```
djhoese, line 224: Just curious, does `np.asanyarray` do the same thing for scalars? This custom function may not be needed.

adybbroe: @djhoese Well it doesn't return the same thing. I think I could replace the call to `array` with `asanyarray`. But to have consistent return values I would keep the function:
```
In [1]: from numpy import where, logical_or, asanyarray, array
In [2]: x = asanyarray(2.3)
In [3]: y = array([2.3])
In [4]: x
Out[4]: array(2.3)
In [5]: y
Out[5]: array([2.3])
In [6]: x.shape, y.shape
Out[6]: ((), (1,))
```

pyspectral/tests/test_reflectance.py
adybbroe: Ok, thanks @pnuu 
But right now I don't manage to make a numpy version with your suggestion above that tests the `numpy.array` function. It keeps testing the built-in function instead!?
I changed your example above to `new=np.array`!

pyspectral/tests/test_reflectance.py
adybbroe: As said below @pnuu I don't see how to patch the numpy `array` function if this doesn't work:
```
with patch('pyspectral.near_infrared_reflectance.array', new=np.array):
    res = get_as_array(1)
    assert res[0] == 1
```
And it doesn't behave as I expected. The code is using the built-in `array` not the Numpy one!
Any ideas?

adybbroe: Ok. So I was fooled. By this debugging:
```
ipdb> array
<built-in function array>
ipdb> array.__module__
'numpy'
```
So, the array patching above that you suggested @pnuu works, is my conclusion.
But now I actually changed the function to always use `asanyarray`
Do you see any problem is that @djhoese or @pnuu ?

djhoese: No, but I'm not really sure everything that gets done with this so it is worth a shot.

adybbroe: Ok, thanks!

PR #131, opened 2021-08-30
Allow more flexible selection of RSRs by band name

simonrp84:
This PR allows users to select a relative spectral response from a wider variety of channel names.
For example, at present Pyspectral allows selection of SEVIRI channels using only `my_rsr['VIS0.6']` or `my_rsr['IR10.8']`. This can be confusing, as it doesn't match the channels names used in other sources such as satpy.
Through this PR, the same RSRs can be selected using multiple band name formats, including both of the above examples.

 - [x] Tests added 
 - [ ] Fully documented


djhoese (commented):
Is the `.rsr` property something that you would say is "public interface"? What do you think about moving the `__getitem__` to the `RSRDataBaseClass` and using `self[something]` instead? If `.rsr` is used everywhere then this is probably a good solution (maybe change to `super().__init__()` in the init though).

adybbroe (commented):
Many thanks @simonrp84 
Could you just add a bit of testing, and fix the stickler complaints?

simonrp84:
@adybbroe I've now added some tests and fixed stickler.

adybbroe (approved):
Thanks a lot @simonrp84 for this. I fixed a few minor flake8 issues and removed some python2 stuff.

simonrp84:
Interesting, pycharm didn't warn me of those issues! Anyway, thanks for fixing. I had left the Python 2 imports as I remove them in my other PR - but I guess it doesn't really matter which PR removes them :)

adybbroe:
> Interesting, pycharm didn't warn me of those issues! Anyway, thanks for fixing. I had left the Python 2 imports as I remove them in my other PR - but I guess it doesn't really matter which PR removes them :)

Yes, thought so, but they stood out shouting at me, so couldn't leave them! :-)

mraspaud (approved):
LGTM!

PR #133, opened 2021-12-06
Refactor rayleigh correction to be more dask friendly

djhoese:
### Update Description

After letting this PR sit for a while I had some other ideas and also started playing around with generating satpy products using a dask distributed scheduler (LocalCluster). I found out that the rayleigh correction can't be used on a distributed dask scheduler as it opens HDF5 files with h5py and these objects can't be serialized. The solution is to either load the numpy arrays into memory and pass them to dask or to open the files inside a `map_blocks` call so the file is only open on the workers and then closed.

This PR uses both. It loads the LUT file initially and gets the rayleigh reflectance and wavelength data. Using this it filters out the specific section of the rayleigh LUT and passes that to a `map_blocks` call. Inside that map blocks function the LUT is loaded again and gets the 3 angle LUT variables which are used for interpolation. There may be some other optimizations that could be made, but this works pretty well so far to make the rayleigh correction dask distributed friendly and saves multiple GBs of memory. See the latest comments below for more information.

### Original Description

I just wanted to see what would happen if I rearranged how the rayleigh code did stuff and if it improved performance. I would consider this negligible, but didn't want to lose the work so here it is. The main difference here is that the LUT file for rayleigh correction (the HDF5 file) is loaded repeatedly for each chunk. This means that dask doesn't have to cache or hold onto the LUT information, but it also means that the workers have to have access to the LUT files. This may not be true in a distributed dask setting.

Original:

image

New:

image

CC @simonrp84 who was interested in this work.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented

### DIFF
--- a/pyspectral/near_infrared_reflectance.py
+++ b/pyspectral/near_infrared_reflectance.py
@@ -40,7 +40,7 @@ except ImportError:
 from pyspectral.solar import (SolarIrradianceSpectrum,
                               TOTAL_IRRADIANCE_SPECTRUM_2000ASTM)
 from pyspectral.utils import BANDNAMES, get_bandname_from_wavelength
-from pyspectral.utils import get_tb2rad_dir
+from pyspectral.utils import _get_tb2rad_dir
 from pyspectral.utils import WAVE_LENGTH
 from pyspectral.radiance_tb_conversion import RadTbConverter
 from pyspectral.config import get_config
@@ -108,7 +108,7 @@ class Calculator(RadTbConverter):
         self.lutfile = None
 
         platform_sensor = platform_name + '-' + instrument
-        tb2rad_dir = get_tb2rad_dir()
+        tb2rad_dir = _get_tb2rad_dir()
         if platform_sensor in options and 'tb2rad_lut_filename' in options[platform_sensor]:
             if isinstance(options[platform_sensor]['tb2rad_lut_filename'], dict):
                 for item in options[platform_sensor]['tb2rad_lut_filename']:
--- a/pyspectral/near_infrared_reflectance.py
+++ b/pyspectral/near_infrared_reflectance.py
@@ -40,7 +40,7 @@ except ImportError:
 from pyspectral.solar import (SolarIrradianceSpectrum,
                               TOTAL_IRRADIANCE_SPECTRUM_2000ASTM)
 from pyspectral.utils import BANDNAMES, get_bandname_from_wavelength
-from pyspectral.utils import TB2RAD_DIR
+from pyspectral.utils import get_tb2rad_dir
 from pyspectral.utils import WAVE_LENGTH
 from pyspectral.radiance_tb_conversion import RadTbConverter
 from pyspectral.config import get_config
@@ -108,6 +108,7 @@ class Calculator(RadTbConverter):
         self.lutfile = None
 
         platform_sensor = platform_name + '-' + instrument
+        tb2rad_dir = get_tb2rad_dir()
         if platform_sensor in options and 'tb2rad_lut_filename' in options[platform_sensor]:
             if isinstance(options[platform_sensor]['tb2rad_lut_filename'], dict):
                 for item in options[platform_sensor]['tb2rad_lut_filename']:
@@ -126,14 +127,14 @@ class Calculator(RadTbConverter):
             if self.lutfile and not os.path.exists(os.path.dirname(self.lutfile)):
                 LOG.warning(
                     "Directory %s does not exist! Check config file", os.path.dirname(self.lutfile))
-                self.lutfile = os.path.join(TB2RAD_DIR, os.path.basename(self.lutfile))
+                self.lutfile = os.path.join(tb2rad_dir, os.path.basename(self.lutfile))
 
         if self.lutfile is None:
             LOG.info("No lut filename available in config file. "
                      "Will generate filename automatically")
             lutname = 'tb2rad_lut_{0}_{1}_{band}'.format(
                 self.platform_name.lower(), self.instrument.lower(), band=self.bandname.lower())
-            self.lutfile = os.path.join(TB2RAD_DIR, lutname + '.npz')
+            self.lutfile = os.path.join(tb2rad_dir, lutname + '.npz')
 
         LOG.info("lut filename: " + str(self.lutfile))
         if not os.path.exists(self.lutfile):
--- a/pyspectral/near_infrared_reflectance.py
+++ b/pyspectral/near_infrared_reflectance.py
@@ -31,6 +31,7 @@ window channel (usually around 11-12 microns).
 
 import os
 import logging
+import tempfile
 import numpy as np
 try:
     from dask.array import where, logical_or, asanyarray, isnan
@@ -40,7 +41,6 @@ except ImportError:
 from pyspectral.solar import (SolarIrradianceSpectrum,
                               TOTAL_IRRADIANCE_SPECTRUM_2000ASTM)
 from pyspectral.utils import BANDNAMES, get_bandname_from_wavelength
-from pyspectral.utils import _get_tb2rad_dir
 from pyspectral.utils import WAVE_LENGTH
 from pyspectral.radiance_tb_conversion import RadTbConverter
 from pyspectral.config import get_config
@@ -108,7 +108,7 @@ class Calculator(RadTbConverter):
         self.lutfile = None
 
         platform_sensor = platform_name + '-' + instrument
-        tb2rad_dir = _get_tb2rad_dir()
+        tb2rad_dir = options.get('tb2rad_dir', tempfile.gettempdir())
         if platform_sensor in options and 'tb2rad_lut_filename' in options[platform_sensor]:
             if isinstance(options[platform_sensor]['tb2rad_lut_filename'], dict):
                 for item in options[platform_sensor]['tb2rad_lut_filename']:
--- a/pyspectral/rayleigh.py
+++ b/pyspectral/rayleigh.py
@@ -186,35 +186,36 @@ class Rayleigh(RayleighConfigBaseClass):
 
         LOG.debug('LUT filename: %s', str(self.reflectance_lut_filename))
 
-    def get_effective_wavelength(self, bandname):
+    def _get_effective_wavelength(self, band_name_or_wavelength):
         """Get the effective wavelength with Rayleigh scattering in mind."""
         try:
             rsr = RelativeSpectralResponse(self.platform_name, self.sensor)
         except OSError:
             LOG.exception(
                 "No spectral responses for this platform and sensor: %s %s", self.platform_name, self.sensor)
-            if isinstance(bandname, (float, int)):
+            if isinstance(band_name_or_wavelength, (float, int)):
                 LOG.warning(
-                    "Effective wavelength is set to the requested band wavelength = %f", bandname)
-                return bandname
+                    "Effective wavelength is set to the requested band wavelength = %f", band_name_or_wavelength)
+                return band_name_or_wavelength
 
             msg = ("Can't get effective wavelength for band %s on platform %s and sensor %s" %
-                   (str(bandname), self.platform_name, self.sensor))
+                   (str(band_name_or_wavelength), self.platform_name, self.sensor))
             raise KeyError(msg)
 
-        if isinstance(bandname, str):
-            bandname = BANDNAMES.get(self.sensor, BANDNAMES['generic']).get(bandname, bandname)
-        elif isinstance(bandname, (float, int)):
-            if not(0.4 < bandname < 0.8):
+        if isinstance(band_name_or_wavelength, str):
+            band_name_or_wavelength = BANDNAMES.get(self.sensor, BANDNAMES['generic']).get(
+                band_name_or_wavelength, band_name_or_wavelength)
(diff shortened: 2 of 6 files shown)


djhoese:
Note the above was generated AHI B01 with sunz_corrected and rayleigh_corrected in satpy but does not include enhancements or writing to disk.

djhoese:
Eh looks like it increases memory usage (and maybe execution time) for a full AHI true_color:

image

pyspectral/tests/test_rayleigh.py
```diff
@@ -115,40 +113,28 @@ def setUp(self):
...
-        exists.return_code = True
-        get_rsr_version.return_code = RSR_DATA_VERSION
-        get_effective_wvl.return_value = self.cwvl
-
+    def test_dask_cliping_angles_with_nans(self):
+        """Test the clipping of angles outside coordinate range - with nan's in input."""
+        from pyspectral.rayleigh import _clip_angles_inside_coordinate_range
         zenith_angle = da.array([79., 69., 32., np.nan])
-        result = self.viirs_rayleigh._clip_angles_inside_coordinate_range(zenith_angle, 2.75)
-
```
djhoese, line 137: This test (and its counterpart in the other class) are a little too bad. Full set of mocks and they were never actually used/needed.

djhoese:
Possibly small increase in execution time but major improvement in memory (~4GB less) for generating an AHI true_color (skipping geotiff creation):

image

pyspectral/rayleigh.py
```diff
@@ -262,80 +263,39 @@ def get_reflectance(self, sun_zenith, sat_zenith, azidiff, bandname, redband=Non
...
 
-        res = clip(nan_to_num(res), 0, 100)
+        res = np.clip(np.nan_to_num(res), 0, 100)
         if compute:
             res = res.compute()
-
         return res
 
+    @staticmethod
+    def _correct_red_by_interpolated_rayleigh_refl(redband, rayleigh_refl):
```
adybbroe, line 295: Maybe reconsider the name of this function?
I didn't understand it in relation to what its doing.
The purpose here is to relax the rayleigh correction in cloudy areas. We use the red band reflection as a proxy for cloud presence. The modelled correction strictly applied only in a cloudfree atmosphere. 
It is hard with appropriate names: Maybe `_relax_rayleigh_reflection_correction_term_where_cloudy`? Or along those lines?

adybbroe (commented):
Many thanks for this! That is much needed. As far as I am able to follow this, it is a nice improvement.
I can try test it here locally when you feel it is ready?

pyspectral/rayleigh.py
djhoese: Ooohhh I think I have always just misunderstood what the point of the `redband` parameter was in this code. I thought this was applying the rayleigh correction to the red band but this is obviously wrong now that I've thought about it for two seconds. I don't mind your suggested long name. I may have just used `_adjust_rayleigh_with_red_band_clouds` or something like that. I can use your name.

One question that this reminds me of though: what is the reflectance variable in the LUT file? It is called `reflectance` in the file, but then called `rayleigh` later and then called `raywl` or something later. So I just started calling it "rayleigh reflectance", but is it a reflectance? Is it a factor? What is it?

pyspectral/rayleigh.py
```diff
@@ -344,46 +304,66 @@ def reduce_rayleigh_highzenith(zenith, rayref, thresh_zen, maxzen, strength):
...
     """Get reflectance LUT.
 
     Read the Look-Up Tables from file with reflectances as a function of
     wavelength, satellite zenith secant, azimuth difference angle, and sun
-    zenith secant
+    zenith secant.
 
     """
-    h5f = h5py.File(filename, 'r')
+    h5f = h5py.File(lut_filename, 'r')
```
mraspaud, line 354: I think it would be cleaner and more exception-proof if we used a context manager here:
```python
with h5py.File(lut_filename, 'r') as h5f:
   ...

```

mraspaud (changes requested):
LGTM, just one comment inline

pyspectral/rayleigh.py
djhoese: So it could have been a fluke, but when I did this on the `_get_rayleigh_wavelength_lut_from_file` function, it crashed with some weird error. Probably because I'm returning the HDF5 objects and by the time they are used the file is closed. I'll double check though.

pyspectral/rayleigh.py
```diff
@@ -262,80 +263,39 @@ def get_reflectance(self, sun_zenith, sat_zenith, azidiff, bandname, redband=Non
...
-                chunks = sun_zenith.chunks if redband is None else redband.chunks
-                res = zeros(shape, chunks=chunks)
-                return res.compute() if compute else res
+        # if the user gave us non-dask arrays we'll use the dask
+        # version of the algorithm but return numpy arrays back
+        compute = HAVE_DASK and not isinstance(sun_zenith, Array)
+
+        rayleigh_refl = _get_wavelength_adjusted_lut_rayleigh_reflectance(self.reflectance_lut_filename, wvl)
+        if rayleigh_refl is None:
+            LOG.warning("Effective wavelength for band %s outside 400-800 nm range!",
```
mraspaud, line 272: Just a flow comment here: wouldn't it be nicer if `_get_wavelength_adjusted_lut_rayleigh_reflectance` raise an error with that message "Effective wavelength for band %s outside 400-800 nm range!" that we catch here and issue a zeroed array in the except clause?

djhoese: Yes. I had tried something like that but decided against it for some reason. I'll try again after some refactoring of the tests.

pyspectral/rayleigh.py
```diff
@@ -344,46 +304,66 @@ def reduce_rayleigh_highzenith(zenith, rayref, thresh_zen, maxzen, strength):
...
-        factor = 1. - strength * where(zenith < thresh_zen, 0, (zenith - thresh_zen) / (maxzen - thresh_zen))
+        factor = 1. - strength * np.where(zenith < thresh_zen, 0, (zenith - thresh_zen) / (maxzen - thresh_zen))
         # For low zenith factor can be greater than one, so we need to clip it into a sensible range.
-        factor = clip(factor, 0, 1)
+        factor = np.clip(factor, 0, 1)
         return rayref * factor
 
 
-def get_reflectance_lut_from_file(filename):
+def _get_zeroed_result(shape, compute, chunks=None):
```
mraspaud, line 313: could `np.zeroes_like` be used here (or the dask counterpart)?

pyspectral/rayleigh.py
djhoese: This has been done now.

djhoese:
@mraspaud @adybbroe I have a few more things I could do, but need your feedback for some of the stuff I just finished. I refactored `get_effective_wavelength` to make it private now as well as some other cleanup. The main thing is that a couple years ago someone made it so `get_reflectance` only ever calls `get_effective_wavelength` with a band name (a string). However, there was a lot of logic left in `get_effective_wavelength` for handling floating point values and getting the band name string by doing a reverse lookup with the RSR data. But again, this logic was never used because of the semi-recent change to make floats never passed to the method. Since no one missed this logic being used, I've removed it. The method now returns two things:

1. The wavelength in nanometers.
2. The band name as a string. If a wavelength is passed this is just a string of the wavelength with `um` appended.

How do you feel about this?

pyspectral/rayleigh.py
djhoese: Ok I think this is done.

pyspectral/rayleigh.py
djhoese: This is kind of done. Turns out the actual condition being checked is not for the 400-800 range, but whether or not the wavelength is within the range in the LUT wavelength variable. I think the higher level code can assume 400-800 though. Let me know what you think.

djhoese:
My understanding of the coverage changes is that some test must have been accidentally downloading the RSRs, but now that I've updated the tests to use fake LUTs and to use a custom config that always says not to download RSRs these download functions aren't getting triggered. Additionally, now that the config loading and directory setup only happen by function call (instead of at import time) some of these functions don't seem to be used (I think).

mraspaud (commented):
A few more small things, but is starts to look good!

pyspectral/rayleigh.py
```diff
@@ -174,216 +172,180 @@ def __init__(self, platform_name, sensor, **kwargs):
...
         """
-        clip_angle = nan_to_num(rad2deg(arccos(1. / zenith_secant_max)))
-        zang = nan_to_num(zenith_angle)
-        return clip(zang, 0, clip_angle)
-
-    def get_reflectance(self, sun_zenith, sat_zenith, azidiff, bandname, redband=None):
-        """Get the reflectance from the three sun-sat angles."""
         # Get wavelength in nm for band:
-        if isinstance(bandname, float):
+        if isinstance(band_name_or_wavelength, (float, int)):
```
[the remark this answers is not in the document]
djhoese: I suppose, but wouldn't that include complex numbers? I've never used `numbers.Number` outside of satpy so I've never really known what it completely represented or how common it was. I can update this to use that.

pyspectral/rayleigh.py
```diff
@@ -24,48 +24,51 @@
...
     da = None
-    map_blocks = None
-    from_array = None
-    Array = None
-    HAVE_DASK = False
 
 from geotiepoints.multilinear import MultilinearInterpolator
 from pyspectral.rsr_reader import RelativeSpectralResponse
-from pyspectral.utils import (INSTRUMENTS, RAYLEIGH_LUT_DIRS,
+from pyspectral.utils import (INSTRUMENTS, _get_rayleigh_lut_dir,
```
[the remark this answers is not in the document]
djhoese: This was something that came up on slack. Can a private function be used in other modules? Maybe private functions can only be used inside the module, and private modules can be used inside the package? I can update this to be public.

pyspectral/rayleigh.py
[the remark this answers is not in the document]
mraspaud: If you're worried about complex numbers, you can use `numbers.Real` if you prefer

pyspectral/rayleigh.py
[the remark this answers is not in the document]
djhoese: Done. I made this one public.

pyspectral/near_infrared_reflectance.py
```diff
@@ -40,7 +40,7 @@
 from pyspectral.solar import (SolarIrradianceSpectrum,
                               TOTAL_IRRADIANCE_SPECTRUM_2000ASTM)
 from pyspectral.utils import BANDNAMES, get_bandname_from_wavelength
-from pyspectral.utils import TB2RAD_DIR
+from pyspectral.utils import _get_tb2rad_dir
```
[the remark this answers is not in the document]
djhoese: This function was actually only used in one spot so I moved the logic to where it was actually being used.

pyspectral/rayleigh.py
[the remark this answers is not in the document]
djhoese: Done.

pyspectral/near_infrared_reflectance.py
```diff
@@ -108,6 +108,7 @@ def __init__(self, platform_name, instrument, band,
         self.lutfile = None
 
         platform_sensor = platform_name + '-' + instrument
+        tb2rad_dir = options.get('tb2rad_dir', tempfile.gettempdir())
```
djhoese, line 111: @adybbroe This tempdir creation as a default differs from the previous implementation in `utils.py` because now a tempdir is created for every instance of this class. In `main` this directory is only created once. Do you know if that is a problem for this Calculator object? I mean, is it a performance problem that the directory changes for every instance?

pyspectral/rayleigh.py
adybbroe: @djhoese It is a reflectance. The data in the LUT file is the simulated clear-sky reflectance with a dark background (none-reflective surface). So, that is mainly the Rayleigh scattering contribution coming from the atmosphere - but it is not exclusively Rayleigh scattering, as the simulations of this signal includes aerosol absorption. Thus, the value from the LUT should be subtracted from the observed reflectance to obtain the (atmosphere) corrected reflectance.

Don't know if @rolle-at-work wants to add something here?

pyspectral/rayleigh.py
adybbroe: Good. Yes, for the LUT we simulated reflectance in the 400 to 800 nm range.

adybbroe:
> @mraspaud @adybbroe I have a few more things I could do, but need your feedback for some of the stuff I just finished. I refactored `get_effective_wavelength` to make it private now as well as some other cleanup. The main thing is that a couple years ago someone made it so `get_reflectance` only ever calls `get_effective_wavelength` with a band name (a string). However, there was a lot of logic left in `get_effective_wavelength` for handling floating point values and getting the band name string by doing a reverse lookup with the RSR data. But again, this logic was never used because of the semi-recent change to make floats never passed to the method. Since no one missed this logic being used, I've removed it. The method now returns two things:
> 
>     1. The wavelength in nanometers.
> 
>     2. The band name as a string. If a wavelength is passed this is just a string of the wavelength with `um` appended.
> 
> 
> How do you feel about this?

I like that, thanks!

PR #134, opened 2021-12-09
Change tested Python versions to 3.8, 3.9 and 3.10

pnuu:
Drop Python 3.7 from the tested Python versions and add Python 3.10.


adybbroe (approved):
LGTM

PR #136, opened 2021-12-14
Fix avhrr instrument naming

### LINKED ISSUE #127: AVHRR instrument naming - allow avhrr-#

```python
# Your code here
from pyspectral.rsr_reader import RelativeSpectralResponse
avhrr3 = RelativeSpectralResponse('NOAA-19', 'avhrr-3')

```
#### Problem description

The above code sample gives a warning. The accepted naming of the sensor is `avhrr/3` and not `avhrr-3`. The former is how it is displayed on WMO Oscar, hence this is the accepted naming in Pyspectral.

It would be fair though to also accept `avhrr-3`. This is also what Satpy uses, as `/` can be problematic to handle in a yaml encoding for instance.

This means Satpy is always raising a warning here. Pyspectral anyhow fixes it so it reads the correct RSR data anyhow, so the warning in Satpy is most disturbing and confusing the user.

Python 3.8
Latest Pyspectral release: 0.10.2

adybbroe:
- [x] Closes #127  
 - [x] Tests added 
 - [ ] Fully documented


doc/conf.py
```diff
@@ -106,7 +106,7 @@ def __getattr__(cls, name):
 
 # General information about the project.
 project = u'Pyspectral'
-copyright = u'2013-2018, PyTroll'
+copyright = u'2013-2021, Pytroll'
```
djhoese, line 109: Depending on how bored you are you could make this use the current year by importing datetime and doing:

```suggestion
copyright = f'2013-{datetime.utcnow():%Y}, Pytroll'
```

adybbroe:
> @adybbroe what is the state of this PR?

Very good question. Will try to find out...

pyspectral/rsr_reader.py
```diff
@@ -126,7 +127,7 @@ def __init__(self, platform_name=None, instrument=None, **kwargs):
         super(RelativeSpectralResponse, self).__init__()
 
         self.platform_name = platform_name
-        self.instrument = instrument
+        self.instrument = AVHRR_INSTRUMENT_NAME.get(instrument, instrument)
```
djhoese, line 130: This seems like an odd place to have something AVHRR specific. Why isn't this handled in `_check_instrument`? Could `_check_instrument` return a new converted instrument instead of nothing like it does now? Oh I see `_check_instrument` is only called if instrument and platform are both provided...that should still work right?

adybbroe: You are right, thanks!
Done some refactoring, also got rid of the `**kwargs` in the class initialization.

pyspectral/utils.py
```diff
@@ -284,13 +286,13 @@ def download_rsr(dest_dir=None, dry_run=False):
     functions from the internet as tarballs, extracts them, then deletes
     the tarball.
 
-    See :func:`pyspectral.rsr_reader.check_and_download` for a "smart" version
+    See: func: `pyspectral.rsr_reader.check_and_download` for a "smart" version
```
djhoese, line 289: Not sure what did this, but this formatting is wrong. `func` is supposed to be surrounded by the colons and right next to the backtick of the function path.

pyspectral/utils.py
```diff
@@ -284,13 +286,13 @@ def download_rsr(dest_dir=None, dry_run=False):
...
     of this process that only downloads the necessary files.
 
     Args:
-        dest_dir (str): Path to put the temporary tarball and extracted RSR
+        dest_dir(str): Path to put the temporary tarball and extracted RSR
```
djhoese, line 293: Also not sure if this was automated or not, but I'm pretty sure it is required/standard that there be a space between the argument name and data type.

doc/conf.py
adybbroe: Ok, yes, I agree. Can we defer this to a more general discussion on how we should handle the headers in all Pytroll code? I rather don't want to change this now in this PR.

pyspectral/utils.py
adybbroe: Really weird! That must have been automated. And be due to the pre-commit stuff.

pyspectral/utils.py
adybbroe, line 289:
```suggestion
    See :func: `pyspectral.rsr_reader.check_and_download` for a "smart" version
```

pyspectral/utils.py
```diff
@@ -284,13 +286,13 @@ def download_rsr(dest_dir=None, dry_run=False):
     functions from the internet as tarballs, extracts them, then deletes
     the tarball.
 
-    See :func:`pyspectral.rsr_reader.check_and_download` for a "smart" version
+    See :func: `pyspectral.rsr_reader.check_and_download` for a "smart" version
     of this process that only downloads the necessary files.
 
     Args:
-        dest_dir (str): Path to put the temporary tarball and extracted RSR
+        dest_dir(str): Path to put the temporary tarball and extracted RSR
```
adybbroe, line 293:
```suggestion
        dest_dir (str): Path to put the temporary tarball and extracted RSR
```

pyspectral/utils.py
```diff
@@ -311,16 +313,16 @@ def download_rsr(dest_dir=None, dry_run=False):
 def download_luts(aerosol_types=None, dry_run=False, aerosol_type=None):
     """Download the luts from internet.
 
-    See :func:`pyspectral.rayleigh.check_and_download` for a "smart" version
+    See: func: `pyspectral.rayleigh.check_and_download` for a "smart" version
```
adybbroe, line 316:
```suggestion
    See :func: `pyspectral.rayleigh.check_and_download` for a "smart" version
```

pyspectral/utils.py
```diff
@@ -311,16 +313,16 @@ def download_rsr(dest_dir=None, dry_run=False):
...
     of this process that only downloads the necessary files.
 
     Args:
-        aerosol_types (Iterable): Aerosol types to download the LUTs for.
-            Defaults to all aerosol types. See :data:`AEROSOL_TYPES` for the
+        aerosol_types(Iterable): Aerosol types to download the LUTs for.
```
adybbroe, line 320:
```suggestion
        aerosol_types (Iterable): Aerosol types to download the LUTs for.
```

pyspectral/utils.py
```diff
@@ -311,16 +313,16 @@ def download_rsr(dest_dir=None, dry_run=False):
...
+            Defaults to all aerosol types. See: data: `AEROSOL_TYPES` for the
```
adybbroe, line 321:
```suggestion
            Defaults to all aerosol types. See :data: `AEROSOL_TYPES` for the
```

pyspectral/utils.py
```diff
@@ -311,16 +313,16 @@ def download_rsr(dest_dir=None, dry_run=False):
...
             full list.
-        dry_run (bool): If True, don't actually download files, only log what
+        dry_run(bool): If True, don't actually download files, only log what
```
adybbroe, line 323:
```suggestion
        dry_run (bool): If True, don't actually download files, only log what
```

pyspectral/utils.py
```diff
@@ -311,16 +313,16 @@ def download_rsr(dest_dir=None, dry_run=False):
...
             URLs would be downloaded. Defaults to False.
-        aerosol_type (str): Deprecated.
+        aerosol_type(str): Deprecated.
```
adybbroe, line 325:
```suggestion
        aerosol_type (str): Deprecated.
```

pyspectral/utils.py
```diff
@@ -461,7 +463,7 @@ def np2str(value):
     """Convert an `numpy.string_` to str.
 
     Args:
-        value (ndarray): scalar or 1-element numpy array to convert
+        value(ndarray): scalar or 1-element numpy array to convert
```
adybbroe, line 466:
```suggestion
        value (ndarray): scalar or 1-element numpy array to convert
```

pyspectral/utils.py
```diff
@@ -284,13 +286,13 @@ def download_rsr(dest_dir=None, dry_run=False):
...
 
-    See :func:`pyspectral.rsr_reader.check_and_download` for a "smart" version
+    See :func: `pyspectral.rsr_reader.check_and_download` for a "smart" version
     of this process that only downloads the necessary files.
 
     Args:
         dest_dir (str): Path to put the temporary tarball and extracted RSR
             files.
-        dry_run (bool): If True, don't actually download files, only log what
+        dry_run(bool): If True, don't actually download files, only log what
```
adybbroe, line 295:
```suggestion
        dry_run (bool): If True, don't actually download files, only log what
```

pyspectral/utils.py
```diff
@@ -311,12 +313,12 @@ def download_rsr(dest_dir=None, dry_run=False):
 def download_luts(aerosol_types=None, dry_run=False, aerosol_type=None):
     """Download the luts from internet.
 
-    See :func:`pyspectral.rayleigh.check_and_download` for a "smart" version
+    See :func: `pyspectral.rayleigh.check_and_download` for a "smart" version
```
adybbroe, line 316:
```suggestion
    See :func:`pyspectral.rayleigh.check_and_download` for a "smart" version
```

pyspectral/utils.py
```diff
@@ -311,12 +313,12 @@ def download_rsr(dest_dir=None, dry_run=False):
...
     of this process that only downloads the necessary files.
 
     Args:
         aerosol_types (Iterable): Aerosol types to download the LUTs for.
-            Defaults to all aerosol types. See :data:`AEROSOL_TYPES` for the
+            Defaults to all aerosol types. See :data: `AEROSOL_TYPES` for the
```
adybbroe, line 321:
```suggestion
            Defaults to all aerosol types. See :data:`AEROSOL_TYPES` for the
```

pyspectral/utils.py
adybbroe: What a mess, I think I have reverted back to how it was now. Thanks for spotting this!

doc/conf.py
djhoese: I'm not sure why it needs to be discussed. Plus this year in this PR is already wrong as it should probably be 2022, right? As for other pytroll code, pyresample and satpy both do the equivalent:

https://github.com/pytroll/pyresample/blob/main/docs/source/conf.py#L67

https://github.com/pytroll/satpy/blob/main/doc/source/conf.py#L116

.pre-commit-config.yaml
```diff
@@ -6,3 +6,13 @@ repos:
     hooks:
     - id: flake8
       additional_dependencies: [flake8-docstrings, flake8-debugger, flake8-bugbear]
+      args: [--max-complexity, "10"]
+-   repo: https://github.com/pycqa/isort
+    rev: 5.10.1
+    hooks:
+      - id: isort
+        language_version: python3
```
djhoese, line 14: If you're adding isort you'll want to include a config for it. Satpy has a pyproject.toml where it configures it:

https://github.com/pytroll/satpy/blob/15be5c1b0c00370d7e2a827c85da25250cad0c1f/pyproject.toml#L8-L14

.pre-commit-config.yaml
```diff
@@ -6,3 +6,13 @@ repos:
...
+ci:
+  # To trigger manually, comment on a pull request with "pre-commit.ci autofix"
+  autofix_prs: false
+  skip: [bandit]
```
djhoese, line 18: This skip isn't needed since you aren't using `bandit`. I think pre-commit.ci will complain about this.

pyspectral/utils.py
djhoese: Yeah, weird. I wouldn't think flake8 would do this and isort shouldn't be touching docstrings.

mraspaud (approved):
LGTM, just one unnecessary line in the pre-commit config.

.pre-commit-config.yaml
```diff
@@ -6,3 +6,12 @@ repos:
     hooks:
     - id: flake8
       additional_dependencies: [flake8-docstrings, flake8-debugger, flake8-bugbear]
+      args: [--max-complexity, "10"]
```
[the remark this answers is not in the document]
adybbroe: Okay, thanks!

adybbroe, line 9:
```suggestion
```

PR #150, opened 2022-06-27
Add AGRI bands to the BANDNAMES dict

simonrp84:
When generating a `true_color` image in satpy I noticed that there's a lot of warnings that the band-specific RSRs cannot be found, which seems to be because in the RSRs the bands are listed as `ch1`, `ch2`, etc whereas satpy uses `C01`, `C02`...

This PR adds AGRI to the `BANDNAMES` dict, enabling the RSRs to be used correctly.


djhoese:
@simonrp84 what is the state of this PR?

adybbroe (approved):
LGTM. I fixed the few stickler issues.

djhoese (approved):

simonrp84:
Thanks for fixing this @adybbroe!

PR #154, opened 2022-09-02
Cleanup RSR and LUT download functions for clearer usage

djhoese:
I was updating some code I have that uses the download functions in pyspectral, but started to get frustrated by lack of docstrings or clear keyword arguments. This PR cleans up docstrings and keyword arguments as well as simplifying some of the logic. It includes one deprecation for the aerosol_type versus aerosol_types keyword argument to be more consistent with the `check_and_download` functions.

 - [x] Tests added 
 - [x] Fully documented

### DIFF
--- a/pyspectral/rsr_reader.py
+++ b/pyspectral/rsr_reader.py
@@ -48,11 +48,15 @@ OSCAR_PLATFORM_NAMES = {'eos-2': 'EOS-Aqua',
 
 
 class RSRDict(dict):
+    """Helper dict-like class to handle multiple names for band keys."""
+
     def __init__(self, instrument=None):
+        """Initialize dict and primary instrument name."""
         self.instrument = instrument
         dict.__init__(self)
 
     def __getitem__(self, key):
+        """Get value either directly or fallback to pre-configured 'standard' names.."""
         try:
             val = dict.__getitem__(self, key)
         except KeyError:
@@ -178,7 +182,6 @@ class RelativeSpectralResponse(RSRDataBaseClass):
 
     def _check_filename_exist(self):
         """Check that the file exist."""
-
         if not os.path.exists(self.filename) or not os.path.isfile(self.filename):
             errmsg = ('pyspectral RSR file does not exist! Filename = ' +
                       str(self.filename))
@@ -334,25 +337,18 @@ class RelativeSpectralResponse(RSRDataBaseClass):
                 self.set_band_central_wavelength_per_detector(h5f, bandname, dname)
 
 
-def check_and_download(**kwargs):
+def check_and_download(dest_dir=None, dry_run=False):
     """Do a check for the version and attempt downloading only if needed."""
-    dry_run = kwargs.get('dry_run', False)
-    dest_dir = kwargs.get('dest_dir', None)
-
     rsr = RSRDataBaseClass()
     if rsr.rsr_data_version_uptodate:
         LOG.info("RSR data already the latest!")
     else:
-        if dest_dir:
-            download_rsr(dest_dir=dest_dir, dry_run=dry_run)
-        else:
-            download_rsr(dry_run=dry_run)
+        download_rsr(dest_dir=dest_dir, dry_run=dry_run)
 
 
 def main():
-    """Main."""
-    modis = RelativeSpectralResponse('EOS-Terra', 'modis')
-    del(modis)
+    """Perform basic sanity check of RSR handling."""
+    RelativeSpectralResponse('EOS-Terra', 'modis')
 
 
 if __name__ == "__main__":
--- a/pyspectral/tests/test_utils.py
+++ b/pyspectral/tests/test_utils.py
@@ -252,7 +252,7 @@ def _create_fake_rsr_tarball_bytes():
     file_obj = BytesIO()
 
     with tarfile.open(mode="w:gz", fileobj=file_obj) as rsr_tar:
-        info = tarfile.TarInfo("PYSPECTRAL_RSR_VERSION")
+        info = tarfile.TarInfo(utils.RSR_DATA_VERSION_FILENAME)
         info.size = len(utils.RSR_DATA_VERSION)
         rsr_tar.addfile(info, BytesIO(utils.RSR_DATA_VERSION.encode("ascii")))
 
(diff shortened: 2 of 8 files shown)


pyspectral/utils.py
```diff
@@ -94,6 +96,7 @@
 ATM_CORRECTION_LUT_VERSION['rayleigh_only'] = {'version': 'v1.0.1',
                                                'filename': 'PYSPECTRAL_ATM_CORR_LUT_RO'}
 
+#: Aerosol types available as downloadable LUTs for rayleigh correction
```
adybbroe, line 99: Could we skip the inline comment is we changed to:
`DOWNLOADABLE_AEROSOL_TYPES_FOR_ATM_CORRECTION`?

Otherwise if you still want the inline comment, change "rayleigh" to "atmospheric". Rayleigh scattering is happening in atmospheric molecules and atoms. Aerosols are mainly absorbing in the visible range. When we speak about rayleigh correction it is the substraction of the enhanced scattering due to atm molecules and atoms. Correcting for aerosol absorption is another atm correction.

pyspectral/rsr_reader.py
```diff
@@ -334,25 +337,18 @@ def get_relative_spectral_responses(self, h5f):
...
+def check_and_download(dest_dir=None, dry_run=False):
     """Do a check for the version and attempt downloading only if needed."""
-    dry_run = kwargs.get('dry_run', False)
-    dest_dir = kwargs.get('dest_dir', None)
-
     rsr = RSRDataBaseClass()
     if rsr.rsr_data_version_uptodate:
         LOG.info("RSR data already the latest!")
     else:
-        if dest_dir:
```
adybbroe, line 346: Is this tested?

pyspectral/utils.py
```diff
@@ -283,8 +296,7 @@ def download_rsr(**kwargs):
 
     config = get_config()
     local_rsr_dir = config.get('rsr_dir')
-    dest_dir = kwargs.get('dest_dir', local_rsr_dir)
-    dry_run = kwargs.get('dry_run', False)
+    dest_dir = dest_dir or local_rsr_dir
```
adybbroe, line 299: Is this tested?

pyspectral/utils.py
```diff
@@ -320,19 +345,9 @@ def download_luts(**kwargs):
...
-
-    if 'aerosol_type' in kwargs:
-        if isinstance(kwargs['aerosol_type'], (list, tuple, set)):
-            aerosol_types = kwargs['aerosol_type']
-        else:
-            aerosol_types = [kwargs['aerosol_type'], ]
-    else:
-        aerosol_types = HTTPS_RAYLEIGH_LUTS.keys()
-
+    aerosol_types = _get_aerosol_types(aerosol_types, aerosol_type)
```
adybbroe, line 348: Is this part tested?

adybbroe (changes requested):
Nice. Thanks a lot for this clean up. Looks good, I only have one minor comment and a wish for a bit more test-coverage if you agree and find it feasible?

pyspectral/utils.py
djhoese: This is considered a "docstring" by sphinx. It was needed so that I could link to this list from the docstring in the related function. If I didn't have this comment here then the constant isn't included in the sphinx docs at all so users can't know what aerosol types are available unless they go into the source code.

pyspectral/rsr_reader.py
djhoese: None of the logic has changed and I didn't write any new tests. If you want me to add tests then I can look into doing that, but otherwise I think I would consider this a refactor (besides the aerosol_type deprecation).

adybbroe: When doing refactoring it is of course preferable to have the coded covered by tests, so we are comfortable nothing breaks when refactoring. And it would be nice to have more test-coverage in general. But it would be a pity if no one dares touching the code because it is already poorly covered, so I just want you to give it a short consideration. I know I am to blame for the lack of test coverage here.

djhoese: I understand the concept, I just assumed things were tested already and didn't look to confirm. Given that this PR isn't urgent for me I can try to add it to my TODO list to try adding tests.

adybbroe (commented):
Really nice with the improved/refactored np2str tests also, thanks.
I wonder if we could use the responses module here?
https://pypi.org/project/responses/
I used it a tiny bit in another project recently, after suggestion by @mraspaud 
I don't want to add to your long To-Do-List, so if you have done some system-testing verifying that the download functions still work the same, we could also leave it here and add an issue to fix better test coverage of the download functionality?

djhoese:
@adybbroe I think I'll add more involved tests still. So far what I added last night was just a placeholder. I'll look at the responses package. If you're OK with it being added then I'm ok with that. I've never used it and I know I did some annoying mocking stuff in Satpy to get something like that working so I'm glad a third-party solution exists.

adybbroe:
> @adybbroe I think I'll add more involved tests still. So far what I added last night was just a placeholder. I'll look at the responses package. If you're OK with it being added then I'm ok with that. I've never used it and I know I did some annoying mocking stuff in Satpy to get something like that working so I'm glad a third-party solution exists.
All good, and thanks @djhoese I am absolutely fine adding another 3rd party package to the dependencies. It will go in the `test_requires` (or what it is called) list of course.

djhoese:
@adybbroe Ok, I think this is ready for re-review. We'll have to wait for the tests to know if code coverage was increased a lot but I'm guessing it is based on my local tests.

pyspectral/tests/test_utils.py
```diff
@@ -181,74 +189,158 @@ def test_get_wave_range(self):
...
+    all_user_warnings = [warn_msg for warn_msg in warns if issubclass(warn_msg.category, UserWarning)]
+    all_matching_warnings = [warn_msg for warn_msg in all_user_warnings if "aerosol_type" in str(warn_msg.message)]
+    assert len(all_matching_warnings) == (1 if exp_warning else 0)
+
+
+def _check_expected_aerosol_files(atypes_to_create, tmp_path):
+    for aerosol_type in utils.AEROSOL_TYPES:
+        atype_fn = tmp_path / aerosol_type / utils.ATM_CORRECTION_LUT_VERSION[aerosol_type]["filename"]
+        if aerosol_type in atypes_to_create:
+            assert os.path.isfile(atype_fn)
```
adybbroe, line 344:
```suggestion
            assert atype_fn.is_file()
```

adybbroe (commented):
LGTM. Thanks for the improved testing!
Only two small suggestions. 
And sorry it took me sol long to get back to you!

adybbroe (approved):

Issue #155, opened 2022-09-16
NOAA-6 spectral responses are wrong for channel 1 & 2


adybbroe:
#### Code Sample, a minimal, complete, and verifiable piece of code

When trying to plot the SW RSR data for NOAA-6 I saw that channel 1 and 2 were off, by what looks like a factor of 10:
  
![rsr_band_0040_1000](https://user-images.githubusercontent.com/169069/190612631-4be23930-f53e-49c6-8fc8-2d4b237a119a.png)

```python
# Your code here
python bin/composite_rsr_plot.py --platform_name NOAA-6 --sensor avhrr -r 0.4 10.0
```

#### Problem description

I found the possible cause for this when digging out the original files from NOAA that were used to generate the internal Pyspectral RSR files. The unit change from Ångström for ch1 and 2 to `nm` for the other channels. `nm` is used for all other platforms and channels for AVHRR/1, so in the conversion script we are using a constant scaling applied to all channels and platforms to go from `nm` to microns!
 
#### Versions of Python, package at hand and relevant dependencies

Thank you for reporting an issue !

djhoese:
This is closed with #157, right?

adybbroe:
> This is closed with #157, right?

Yes, right, fixed that now.

PR #156, opened 2022-09-27
Add support for NOAA-21 VIIRS

adybbroe:
This PR adds support for the JPSS-2 (NOAA-21) VIIRS RSR data.
The actual pyspectral formattet hdf5 files are not yet uploaded to zenodo, so the version pointing to the RSR files remains as before. The RSR file version will be updated in another PR. See #155 

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented

### DIFF
--- a/rsr_convert_scripts/viirs_rsr.py
+++ b/rsr_convert_scripts/viirs_rsr.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 #
-# Copyright (c) 2013-2020 Pytroll developers
+# Copyright (c) 2013-2022 Pytroll developers
 #
 # Author(s):
 #
@@ -21,9 +21,11 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-"""Interface to VIIRS relative spectral responses"""
+"""Interface to read original VIIRS relative spectral responses and write to hdf5 file."""
 
 import os
+import sys
+import h5py
 import numpy as np
 from pyspectral.utils import get_central_wave
 from pyspectral.config import get_config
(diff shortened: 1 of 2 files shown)


djhoese:
Now that #157 is done, does the URL in this PR need to be updated? Or does a new upload to zenodo need to happen?

rsr_convert_scripts/viirs_rsr.py
```diff
@@ -263,4 +254,6 @@ def main():
 
 if __name__ == "__main__":
     LOG = logging.getLogger('viirs_rsr')
```
djhoese, line 256: I just noticed, this logger is already created at the top of the script isn't it? Can this be removed?

rsr_convert_scripts/viirs_rsr.py
```diff
@@ -263,4 +254,6 @@ def main():
...
-    main()
+    platform_name = "NOAA-21"
+    # platform_name = "Suomi-NPP"
```
djhoese, line 258: Remove unused code?

djhoese (approved):
Really small cleanup comments, but otherwise looks good.

adybbroe:
> Now that #157 is done, does the URL in this PR need to be updated? Or does a new upload to zenodo need to happen?

I have already uploaded a new version with NOAA-6 and JPSS-2 updates to Zenodo a few days ago. But you are absolutely right the URL ought to be updated. I merge main into this one.

rsr_convert_scripts/viirs_rsr.py
adybbroe: Yes, right, thanks!

rsr_convert_scripts/viirs_rsr.py
adybbroe: Done

PR #158, opened 2022-09-28
Fix RSR band name retrieval for unconfigured sensors

djhoese:
~This adds the missing configuration for mapping MERSI-2 band names to the names used in the RSR files. It is basically "ruined" by the generic mapping mapping numeric band names to `ch{:d}`. This isn't the case for MERSI-2 which uses the same band name.~

It turns out MERSI-2 does use the `chN` names in the RSR HDF5 files, but since it wasn't configured an `if` statement in the RSR Reader was failing with a KeyError before it was supposed to. This PR should fix this so that the "generic" band names are used as a fallback.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented

### DIFF
--- a/pyspectral/bandnames.py
+++ b/pyspectral/bandnames.py
@@ -195,8 +195,6 @@ BANDNAMES['slstr'] = {'S1': 'ch1',
                       'F2': 'ch8',
                       }
 
-BANDNAMES['mersi-2'] = {str(chan_num): str(chan_num) for chan_num in range(26)}
-
 BANDNAMES['VII'] = {'vii_443': 'ch1',
                     'vii_555': 'ch2',
                     'vii_668': 'ch3',
--- a/pyspectral/bandnames.py
+++ b/pyspectral/bandnames.py
@@ -195,6 +195,8 @@ BANDNAMES['slstr'] = {'S1': 'ch1',
                       'F2': 'ch8',
                       }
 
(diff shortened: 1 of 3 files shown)


djhoese:
I manually checked the HDF5 file and it seems like it is using the `chN` format of variable names, but with or without this PR I'm still getting these warnings:

```
WARNING  : Could not get the reflectance correction using band name: 1
WARNING  : Will try use the wavelength, however, this may be ambiguous!
WARNING  : A wavelength is provided instead of band name - disregard the relative spectral responses and assume it is the effective wavelength: 0.470000 (micro meter)
```

adybbroe:
> I manually checked the HDF5 file and it seems like it is using the `chN` format of variable names, but with or without this PR I'm still getting these warnings:
> 
> ```
> WARNING  : Could not get the reflectance correction using band name: 1
> WARNING  : Will try use the wavelength, however, this may be ambiguous!
> WARNING  : A wavelength is provided instead of band name - disregard the relative spectral responses and assume it is the effective wavelength: 0.470000 (micro meter)
> ```

@djhoese 
This is from Satpy right? Just to be clear. We should change a `logger.warning`to a `logger.exception` instead I think!
That would help quite a bit in debugging Pyspectral. wouldn't it?

PR #159, opened 2022-10-06
Revert np.nan_to_num usage on rayleigh correction array

djhoese:
As discussed on slack, the usage of `np.nan_to_num` that is being changed in this PR was probably originally added with the idea that "if the correction calculations went wrong then we don't want to mask pixels, we just want to *not* correct them". The problem with this logic in some recent real world cases is that it is more "jarring" to see uncorrected pixels next to corrected pixels instead of black/transparent/masked pixels. See this MERSI-2 case where the angles at 1000m resolution do not go as far west as the band data pixels at 250m (for some reason) and this results in a 0 correction for these 5 or so pixels on the left side of the swath:

image

If we use this PR's changes (and the way it was before PR #140) then these pixels are simply masked away as invalid pixels:

image

Note: The gray background in the first image was me playing around with the fill value.

I will work on making an issue to further investigate `np.nan_to_num`'s usage in Rayleigh correction at terminator regions.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented

### DIFF
--- a/pyspectral/rayleigh.py
+++ b/pyspectral/rayleigh.py
@@ -267,7 +267,7 @@ class Rayleigh(RayleighConfigBaseClass):
                                              dtype=res.dtype,
                                              chunks=getattr(res, "chunks", None))
 
-        res = np.clip(np.nan_to_num(res), 0, 100)
+        res = np.clip(res, 0, 100)
         if compute:
             res = res.compute()
         return res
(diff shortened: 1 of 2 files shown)


adybbroe (approved):
Many thanks for your analysis and the proposed changes. LGTM

PR #171, opened 2022-11-10
Update for FY-4B

simonrp84:
The FY-4B satellite flies with the AGRI and GHI instruments. This PR Updates pyspectral to support the FY-4B/AGRI relative spectral response functions and also alters the code to accept a satellite with multiple sensors onboard (currently, only one is supported).

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/pyspectral/rayleigh.py
+++ b/pyspectral/rayleigh.py
@@ -153,7 +153,7 @@ class Rayleigh(RayleighConfigBaseClass):
         # Try to fix instrument naming
         instr = INSTRUMENTS.get(platform_name, sensor)
         if instr != sensor:
-            if type(instr) is list:
+            if isinstance(instr, list):
                 if sensor not in instr:
                     sensor = instr[0]
                     LOG.warning("Inconsistent sensor/satellite input - " +
--- a/pyspectral/rayleigh.py
+++ b/pyspectral/rayleigh.py
@@ -150,12 +150,19 @@ class Rayleigh(RayleighConfigBaseClass):
         self.sensor = sensor
         self.coeff_filename = None
 
-        # Try fix instrument naming
+        # Try to fix instrument naming
         instr = INSTRUMENTS.get(platform_name, sensor)
         if instr != sensor:
-            sensor = instr
-            LOG.warning("Inconsistent sensor/satellite input - " +
-                        "sensor set to %s", sensor)
+            if type(instr) is list:
+                if sensor not in instr:
+                    sensor = instr[0]
+                    print(sensor, instr)
+                    LOG.warning("Inconsistent sensor/satellite input - " +
+                                "sensor set to %s", sensor)
+            else:
+                sensor = instr
+                LOG.warning("Inconsistent sensor/satellite input - " +
+                            "sensor set to %s", sensor)
 
         self.sensor = sensor.replace('/', '')
 
--- a/pyspectral/rayleigh.py
+++ b/pyspectral/rayleigh.py
@@ -156,7 +156,6 @@ class Rayleigh(RayleighConfigBaseClass):
             if type(instr) is list:
                 if sensor not in instr:
                     sensor = instr[0]
-                    print(sensor, instr)
                     LOG.warning("Inconsistent sensor/satellite input - " +
                                 "sensor set to %s", sensor)
             else:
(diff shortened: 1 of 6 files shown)


pyspectral/rayleigh.py
```diff
@@ -150,12 +150,19 @@ def __init__(self, platform_name, sensor, **kwargs):
...
         self.coeff_filename = None
 
-        # Try fix instrument naming
+        # Try to fix instrument naming
         instr = INSTRUMENTS.get(platform_name, sensor)
         if instr != sensor:
-            sensor = instr
-            LOG.warning("Inconsistent sensor/satellite input - " +
-                        "sensor set to %s", sensor)
+            if type(instr) is list:
```
djhoese, line 156:
```suggestion
            if isinstance(instr, list):
```

pyspectral/utils.py
```diff
@@ -550,5 +551,11 @@ def are_instruments_identical(name1, name2):
         return True
     if name1 == INSTRUMENT_TRANSLATION_DASH2SLASH.get(name2):
         return True
+    if type(name1) is list and type(name2) is str:
+        if name2 in name1:
+            return True
+    if type(name2) is list and type(name1) is str:
+        if name1 in name2:
+            return True
```
djhoese, line 559: Same for these type checks.

pyspectral/utils.py
```diff
@@ -550,5 +551,11 @@ def are_instruments_identical(name1, name2):
         return True
     if name1 == INSTRUMENT_TRANSLATION_DASH2SLASH.get(name2):
         return True
+    if type(name1) is list and type(name2) is str:
```
adybbroe, line 554: I think it will work, but the function name is not really representing what this does anymore. Could/should this be refactored a bit to handle both cases (lists versus name as str)?
At least the doc string needs to explain

rsr_convert_scripts/agri_rsr.py
```diff
@@ -17,9 +17,9 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-"""Read the FY-4A AGRI relative spectral responses.
+"""Read the FY-4A/B AGRI relative spectral responses.
 
-Data from http://fy4.nsmc.org.cn/portal/cn/fycv/srf.html
+Data from http://satellite.nsmc.org.cn/PortalSite/StaticContent/DocumentDownload.aspx?TypeID=590
```
adybbroe, line 22: Should we regenerate the 4A AGRI RSR, to use this (new) url?

adybbroe (changes requested):
Nice, many thanks for adding this!
I have some comments and requests if you could consider?

pyspectral/utils.py
simonrp84: Hopefully I have resolved this now!

pyspectral/tests/test_rayleigh.py
```diff
@@ -280,6 +280,17 @@ def test_rayleigh_reduction(self, fake_lut_hdf5):
...
+    def test_rayleigh_getname(self):
+        """Test logic for Rayleigh instrument selection."""
+        rayl = _create_rayleigh(platform='FY-4A')
+        assert rayl.sensor == 'agri'
+        rayl = _create_rayleigh(platform='FY-4B')
+        assert rayl.sensor == 'agri'
+        rayl = _create_rayleigh(platform='FY-4B', sensor='ghi')
+        assert rayl.sensor == 'ghi'
+        rayl = _create_rayleigh(platform='FY-4B', sensor='nosensor')
+        assert rayl.sensor == 'agri'
```
djhoese, line 292: Why does a non-existent sensor return `agri`?

simonrp84: I've updated this now, so if the available sensors is a `list` but the user does not explicitly state which sensor they want then a `ValueError` will be raised.

adybbroe (approved):
LGTM

adybbroe:
So @simonrp84 as discussed on Slack you fix the RSR responses for AGRI FY-4 A&B so that they are all between 0 and 1 before we merge this PR, ok?

PR #174, opened 2022-11-17
Fix bug getting bandname from RSR data

### LINKED ISSUE #172: Bug plotting spectral responses with MODIS

```python
# Your code here
python bin/composite_rsr_plot.py --platform_name Sentinel-3B EOS-Terra --sensor olci modis -r 0.4 0.8
```
#### Problem description

This gives an error, see below. If I remove the OLCI it works fine, and I can add Meteosat FCI and SNPP VIIRS together with modis, but when adding OLCI it fails like below. Strange!
 
[this should also explain **why** the current behaviour is a problem and why the 
expected output is a better solution.]

#### Expected Output

A plot file in png should be generated. Here with only olci:
rsr_band_0040_0080

```
Traceback (most recent call last):
  File "/home/a000680/usr/src/forks/pyspectral/pyspectral/rsr_reader.py", line 64, in __getitem__
    val = dict.__getitem__(self, key)
KeyError: 'ch8'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/a000680/usr/src/forks/pyspectral/bin/composite_rsr_plot.py", line 200, in <module>
    plt = plot_band(plt, b__, rsr,
  File "/home/a000680/usr/src/forks/pyspectral/bin/composite_rsr_plot.py", line 41, in plot_band
    detectors = rsr_obj.rsr[band_name].keys()
  File "/home/a000680/usr/src/forks/pyspectral/pyspectral/rsr_reader.py", line 71, in __getitem__
    raise KeyError(f'Band not found in RSR for {self.instrument}: {key}')
KeyError: 'Band not found in RSR for modis: ch8'

```

Python 3.10 and Pyspectral v0.12.2

adybbroe:
Fix a bug getting the correct band name from the RSR data object.

 - [x] Closes #172  
 - [x] Tests added 
 - [ ] Fully documented


bin/composite_rsr_plot.py
```diff
@@ -149,6 +149,8 @@ def get_arguments():
 
     for platform in platform_names:
         for sensor in sensors:
+            if platform in INSTRUMENTS and sensor not in INSTRUMENTS.get(platform):
+                continue
```
djhoese, line 153: This if statement seems out of place. Does this mean that it will silently not plot any sensors that it doesn't know about? Since that is coming from the user shouldn't that be an error or at least a warning?

adybbroe: It is perhaps not really beautiful, but the user gives a set of platform names and instrument names, so for instance:
`--platform_name Sentinel-3B EOS-Terra --sensor olci modis`

and then here we want to avoid asking for `Terra` and `olci`.

adybbroe: But you are right, I should perhaps split so first checking that Pyspectral actually supports the platform...

adybbroe: The bug in issue #174 was triggered by asking for EOS-Terra and olci! ;-)
Pyspectral has and is still trying to corect what you are trying to do like here:
```
In [3]: rsr = RelativeSpectralResponse('EOS-Terra', 'olci')
Inconsistent instrument/satellite input - instrument set to modis
rsr data may not be up to date: /home/a000680/data/pyspectral/rsr_modis_EOS-Terra.h5
```

Which can be debated of course. But I am not changing that behaviour in this PR.
What I am correcting is that when asking for the channel number you would now use the method, rather than the corresponding function where you before entered with `sensor=olci` and with the (corrected) rsr now valid for Terra modis. That then gave an error. Now it is consistent and gives the correct modis band name:
```
In [4]: rsr.get_bandname_from_wavelength(3.7)
Out[4]: '20'
```

djhoese: Sounds good.

pnuu (approved):
LGTM

PR #200, opened 2023-10-18
Add RSR reader for the EPIC instrument onboard DSCOVR

simonrp84:
This PR adds a conversion script for spectral response functions for the EPIC instrument aboard the DSCOVR satellite. 

It converts the excel file with RSR data into an HDF5 file suitable for pyspectral.


rsr_convert_scripts/epic_reader.py
adybbroe, line 1: Looks fine. The only thing is that the responses are not normalized. I saw that from your HDF5 file you provided. However, right now I can't find another example where we actively normalize the responses if the originals were not. Most often they are. Also MetImage has non-normalized RSR at the moment. So we can leave this for another PR.

I generally think we should make sure all responses are normalized (to 1). It makes plotting various instrument RSRs on the same plot more convenient if nothing else.

adybbroe (approved):
LGTM! Many thanks. Just one comment that I also want your view on?
Maybe @rolle-at-work , @mraspaud and others have an opinion as well here?

adybbroe (changes requested):
We need changing the name of the hdf5 file as said...
I am testing off line. Will come back

rsr_convert_scripts/epic_reader.py
rolle-at-work: Finally managed to login here again...
Yes I had some problems before with non-normalized RSR's. So before I used an RSR I make sure it is normalized and in correct order. 
But it is a good thing to protect others from bad experience by normaliszing RSRs directly.

simonrp84:
OK, have done all the above - should be better now :)

rsr_convert_scripts/epic_reader.py
simonrp84: Done!

adybbroe: Looks good, thanks!

adybbroe (approved):
LGTM. I have updated Zenodo and the link and version in Pyspectral. (Unfortunately forgot to state the right version (1.2.4) on Zenodo, so now it says version 30. But should not have any affect as long as the link in `utils.py` is correct.

PR #202, opened 2023-10-30
Update the solar spectrum code to use the included reference spectra

simonrp84:
At present, the `SolarIrradianceSpectrum` class requires the user to enter the filename to the solar spectra they wish to use.
However, `Pyspectral` comes with a data file containing a solar spectra (2000 ASTM Standard Extraterrestrial Spectrum Reference E-490-00).
This PR updates the class to use the included spectra by default (i.e: In the case that the user does not supply their own via the `filename` argument. I have also updated the `__init__` docstring to make clear that this spectra is used by default.

### DIFF
--- a/pyspectral/solar.py
+++ b/pyspectral/solar.py
@@ -48,14 +48,20 @@ class SolarIrradianceSpectrum(object):
     in units of W/m^2/micron
     """
 
-    def __init__(self, filename, **options):
+    def __init__(self, filename=TOTAL_IRRADIANCE_SPECTRUM_2000ASTM, **options):
         """Initialize the top of atmosphere solar irradiance spectrum object from file.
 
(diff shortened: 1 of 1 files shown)


djhoese:
I'm looking at the use of `pkg_resources` here in my #207 PR and noticed that in a previous PR where I modified `MANIFEST.in` that this data file is no longer included in the source distribution. I'll need to fix this.

djhoese (changes requested):
This PR should update all the files in `doc/` that refer to this variable and use it. The code examples should be much simpler and maybe a note could be added to one or more of these docs pages (I see 3 files using it) to mention what specific file is being used.

djhoese:
Ah nevermind my second to last comment. The files are included!

PR #207, opened 2023-11-27
Remove pkg_resources usage in config handling

djhoese:
`pkg_resources` is deprecated. This removes the usage while also moving the access of the `PSP_CONFIG_FILE` environment variable from an import-time operation to a runtime operation inside the config function call.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


adybbroe (approved):
LGTM Thanks for this!

PR #224, opened 2024-05-03
Add support and RSR converter for the GOCI2 instrument aboard GK-2B

simonrp84:
This PR adds a script to generate the spectral response function data for the GOCI-2 instrument aboard GeoKompsat-2B.
The RSR raw data has been taken from the official website and the generated data file for pyspectral is available here:
[rsr_goci2_GK-2B.zip](https://github.com/pytroll/pyspectral/files/15196965/rsr_goci2_GK-2B.zip).

I have also edited a string in the YAML file to make it less clear that it was me who added EPIC support ;-)

### DIFF
--- a/pyspectral/etc/pyspectral.yaml
+++ b/pyspectral/etc/pyspectral.yaml
@@ -16,7 +16,7 @@
 download_from_internet: True
 
 # Everything below this line should not need to be changed!
-# Changes may be done if you want to change the name of the radiance<->Tb LUT filrnames,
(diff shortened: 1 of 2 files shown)


simonrp84:
The 3.12 unstable tests are failing, but that looks like it's to do with numpy 2 rather than anything to do with this PR.

simonrp84:
> LGTM. Though I fail fetching the original SRFs?

I've now also added a direct link to the SRFs in addition to the (more stable) link to the website that hosts the SRFs.

PR #228, opened 2024-06-27
Fix scipy 1.14 compatibility and trapz usage

### LINKED ISSUE #227: Pyspectral 0.13.1 not compatible with SciPy 1.14.x -> update dependencies

Sorry for not being more verbose, I'm short on time.

```python
AttributeError: module 'scipy.integrate' has no attribute 'trapz'
...
File "/usr/local/lib/python3.11/site-packages/pyspectral/radiance_tb_conversion.py", line 224, in tb2radiance
    radiance = integrate.trapz(planck, self.wavelength_or_wavenumber) / self.rsr_integral
               ^^^^^^^^^^^^^^^
```

#### Problem description

Deprecated method `scipy.integrate.trapz`  removed in SciPy 1.14.0, see https://github.com/scipy/scipy/issues/20486 leads to above exception.

#### Solution suggestions

- Bugfix version of pyspectral with requirement.txt containing
`scipy>=0.14, < 1.14` (or so)
- and/of next major version using `scipy.intergrate.trapezoid`
- check `scipy.integrate` attributes whether `trapz` or `trapezoid` is available

djhoese:
CC @arcanerr

The scipy.integrate.trapz function is deprecated. This adds a try/except around importing trapezoid versus trapz. I'm not sure how long we want to keep this workaround as `trapezoid` has been around since at least January of 2023 from what I can tell in git.

 - [x] Closes #227  
 - [ ] Tests added 
 - [ ] Fully documented

### DIFF
--- a/pyspectral/blackbody.py
+++ b/pyspectral/blackbody.py
@@ -54,6 +54,9 @@ def blackbody_rad2temp(wavelength, radiance):
         The derived temperature in Kelvin.
 
     """
+    if getattr(wavelength, "dtype", None) != radiance.dtype:
+        # avoid a wavelength numpy scalar upcasting radiances (ex. 32-bit to 64-bit float)
+        wavelength = radiance.dtype.type(wavelength)
     with np.errstate(invalid='ignore'):
         return PLANCK_C1 / (wavelength * np.log(PLANCK_C2 / (radiance * wavelength**5) + 1.0))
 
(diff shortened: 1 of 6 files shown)


djhoese:
FYI for you @adybbroe when you find time to review this: A lot of the test changes are because numpy 2's `np.testing.assert_allclose` now checks dtype and asserts that dtypes are the same (or at least as far as I can tell). So a lot of the numpy-based tests would use 64-bit float input arrays and then compare against a 32-bit float expected result array. This used to work fine, but now numpy 2 compares about comparing 32-bit with 64-bit. Otherwise, there are just a few spots where numpy 2 now upcasts things where it didn't before.

adybbroe (approved):
LGTM. Nice with the more compact tests!
I was actually not aware that trapz was deprecated. I think we keep the try-except for now.

PR #231, opened 2024-07-15
Updated the file name for the FY-3F Mersi-3 RSRs

adybbroe:
The purpose of this PR is to handle the fact that Pyspectral expects the RSR filename for Mersi-3 FY-3F to be `rsr_mersi3_FY-3F.h5` rather than `rsr_mersi-3_FY-3F.h5`

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


simonrp84 (approved):
LGTM!

PR #233, opened 2024-08-14
Replace deprecated numpy trapezoid/trapz usage with scipy trapezoid

### LINKED ISSUE #229: Deprecate old scipy and remove use of trapz (numpy and scipy)
Modern versions of scipy have a `scipy.integrate.trapezoid` function that replaces the deprecated `trapz` function. Additionally numpy 2.0 deprecates (but I think still functions) their `np.trapz` function in favor of `np.trapezoid`. Note that as far as I can tell these are the same exact function in numpy and scipy, scipy just wraps the numpy function.

So we need to:

1. Define the minimum version of scipy supported and make sure it has `scipy.integrate.trapezoid`.
2. Remove try/except in `pyspectral/radiance_tb_conversion.py`
3. Replace all `np.trapz` usage with `np.trapezoid` I suppose (note the numpy docs search never seems to find.

djhoese:
This PR completes the removal of deprecated trapz/np.trapezoid usage. It also sets the lower bound of the scipy dependency to 1.6.0 where trapz was renamed to trapezoid.

 - [x] Closes #229 
 - [ ] Tests added 
 - [ ] Fully documented 


djhoese:
Adam and I had talked about the distutils change on slack and needing to release it. This is a similar change where numpy and scipy have both deprecated the "trapz" function in favor of "trapezoid" naming, but numpy is also suggesting people import and use scipy directly (one referenced the other, but my understanding it is all just scipy now).

PR #240, opened 2024-10-27
Fix rsr plotting to allow for similar (overlapping) bands

adybbroe:
Fix the script to plot spectral response functions, allowing for similar (overlapping) bands - and don't just ignore all others than the first found.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


adybbroe:
For instance, now you can do:
```
%> python bin/composite_rsr_plot.py --platform_name Meteosat-11 Meteosat-12 NOAA-21 --sensor seviri fci viirs -r 10.5 12.5 -v --exclude_bandnames I5
```
Which will produce this plot:


Leaving out the last `--exclude_bandnames` option above would have resultet in this plot:

djhoese:
Maybe it's just me but those quicklooks aren't working. They still say "Uploading ...png...". Did you maybe close the tab before they were fully uploaded?

bin/composite_rsr_plot.py
```diff
@@ -103,6 +103,11 @@ def get_arguments():
                        help="The wavelength range for the plot",
                        default=[None, None], type=float)
 
+    parser.add_argument("--exclude_bandnames", nargs='*',
+                        default=[],
+                        required=False,
```
djhoese, line 108: `required` is not necessary here. The default is `False`. Actually `type=str` is also not necessary as `str` is the default type.

simonrp84 (approved):
LGTM! I'm not expert enough to figure out how to do what Dave asks but they appear like reasonable suggestions if you have time. If not, happy to see it merged as-is.

adybbroe:
> Maybe it's just me but those quicklooks aren't working. They still say "Uploading ...png...". Did you maybe close the tab before they were fully uploaded?

You were right, sorry, should work now

djhoese:
@adybbroe Can you fix the title of the PR. It got cut off from your commit message it seems.

For your quicklooks, yes I see them, but now I'm not sure what you are trying to show. Prior to this PR the code would just ignore I5? Or would it fail because two wavelengths matched for a single VIIRS instrument?

And now? Now you don't need the exclude I5 because it can be plotted without issue?

adybbroe:
> Small change requested. I don't understand the code before or after this PR so I'm not sure I should give my opinion too strongly. The code "smells" like there should be a simpler way, but if it works for you then 👍.
> 
> Possible simplification: Could you have a function that gets all the bands needed to be plotted and would take the wavelength and band information in as input and produce a list of bands?

I believe you have a sharp nose! I think you might be right. Will think another day on this, and either start simplify, or merge and make an issue, or just merge :-)

PR #258, opened 2025-09-26
Add RSR reader for SLSTR in text format from NWP SAF

simonrp84:
This PR adds a new RSR conversion script for the Sentinel-3/SLSTR instrument using SRFs retrieved from the NWP SAF spectral library.
The existing reader only works with data in netCDF format downloaded from the sentiwiki page. However, this page has not yet been updated with the SRFs for the -C and -D units of SLSTR...but the SRFs for these units are available via the NWP SAF.

Also, this PR corrects the link in the existing netCDF reader to point to the correct SRF download location.


simonrp84:
pre-commit.ci autofix

PR #259, opened 2025-09-26
Add two helper functions for calculating bandwidth

simon-sat:
The bandwidth of a satellite sensor is the wavelength range over which it is sensitive to a particular level of signal. This is typically defined in one of two ways:
1) The full width half maximum: The total range of wavelengths over which the spectral response function is greater than half of the maximum response.
2) The integrated energy response: The minimum and maximum wavelengths over which a sensor is sensitive to more than a given threshold of the integrated energy measured by the sensor.

This PR adds utility functions to compute these two parameters. It's still a draft as I need to add documentation and tests.

 - [x] Tests added 
 - [x] Fully documented


djhoese (approved):
Not going to pretend everything but the basics look good. Thanks.

pnuu (commented):
The integrated energy test could use parametrization, but not a show stopper.

mraspaud (approved):
I agree with @pnuu 's comment, but otherwise LGTM

PR #264, opened 2025-10-17
Refactor near-infrared reflectance calculator

pnuu:
This PR does some refactoring to `pyspectral.near_infrared_reflectance` module.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/pyspectral/near_infrared_reflectance.py
+++ b/pyspectral/near_infrared_reflectance.py
@@ -63,7 +63,7 @@ class Calculator(RadTbConverter):
     """
 
     def __init__(self, platform_name, instrument, band,
-                 detector='det-1', wavespace=WAVE_LENGTH,
+                 detector="det-1", wavespace=WAVE_LENGTH,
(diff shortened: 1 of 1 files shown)


adybbroe:
Thanks @pnuu nice work. Haven't carefully checked everything yet, it is some re-arrangements, so just wonder if you are confident with the test coverage of all this?

Also, perhaps you saw there was some discussion recently about the "nir" and "near infrared" here, as it is a bit mis-leading and strictly not true, as we are operating with what is normally termed the mid-wave IR range. SO, should this be part of this re-factoring, or rather another one?

I guess that changing the name of functions will break compatibility in Satpy for instance, so should probably go in another PR...

djhoese (approved):
I will be honest and admit that I didn't look at every little change, but I just ran the tests locally with a rebased version of this PR and it all passed so...it's probably fine. Thanks!

PR #271, opened 2026-01-13
Fix Zenodo RSR download by adding User-Agent and HTTP checks

albertbrotzer:
Zenodo seems to block automated downloads that do not provide a proper User-Agent header. As a result, pyspectral may silently download a 403 HTML page instead of the RSR tarball, leading to tarfile read errors.

This commit is meant to provide a minimal fix. It
- adds a proper User-Agent header to the Zenodo download request
- enables streaming downloads
- fails early on HTTP errors instead of extracting invalid files

The issue can be reproduced via Satpy when generating the `day_microphysics` composite for EUMETSAT sensors.

If the pyspectral RSR cache is empty, the first access triggers a download from Zenodo, which currently fails with HTTP 403
when no User-Agent is set.

### DIFF
--- a/pyspectral/utils.py
+++ b/pyspectral/utils.py
@@ -411,6 +411,7 @@ HEADERS = {
     "Accept": "*/*",
 }
 
+
 def _download_tarball_and_extract(tarball_url, local_pathname, extract_dir):
     chunk_size = 1024 * 1024  # 1 MB
 
--- a/pyspectral/utils.py
+++ b/pyspectral/utils.py
@@ -407,7 +407,7 @@ def _get_aerosol_types(aerosol_types, aerosol_type):
 
 
 HEADERS = {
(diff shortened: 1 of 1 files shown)


djhoese (changes requested):
Thank you! We've been trying to figure this out, but starting with workarounds like testing utilities to mock the downloads (see #269). I had a few questions that I made comments for inline. Otherwise, is there a publication/blog post by zenodo that defines this requirement and possibly when the change happened?

djhoese (approved):
Looks good to me. Again, thanks so much for figuring this out and fixing it.

@adybbroe or @mraspaud any last reviews before merge?

pnuu (approved):

mraspaud (approved):
LGTM, and I appreciate the small clean ups also!

PR #275, opened 2026-01-31
Fix testing utilities using strings as Path

djhoese:
In #262 I added some creation of temporary directories if the user didn't provide one themselves (ex. from `tmp_path` in a pytest test). However, I never cast the location of this temporary directory to a `Path` object so later use of it failed. This PR fixes that and adds some tests specifically for the testing utilities.

Side note: I kind of want to replace the way I create these temporary directories with a decorator on the functions where they check if the user provided a temp dir and if not they create one and use it as a context manager. It makes it a little hard to follow the code but also makes it much cleaner and I can reuse the decorator on all the functions that need it.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 


pnuu (approved):
LGTM

PR #276, opened 2026-02-06
Fix RSR downloads not downloading on-demand

djhoese:
Major breaking change in #269. Due to incorrect test driven development ordering, I prevented and guarded against RSR files being downloaded, but there were no tests making sure they would still get downloaded. The main cause was the reuse of the argument names in the RSR reader. This PR splits the usage into "what the user passed" and "what we determined we should use after sanitizing". I also moved more functionality from the RSR tests to the testing utilities.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/pyspectral/rsr_reader.py
+++ b/pyspectral/rsr_reader.py
@@ -124,7 +124,6 @@ class _RSRDataBase:
             updated was needed.
 
         """
-        print(self.rsr_data_version, self.rsr_data_version_uptodate, self.do_download)
         if self.rsr_data_version_uptodate:
             return False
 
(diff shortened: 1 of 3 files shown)


mraspaud (approved):
This LGTM, and can be merged I think.

Some thoughts to ponder maybe for future development:
We now have a test that downloads from zenodo, and we've seen how unstable this can be. So should we have a special ci item that i allowed to go wrong? To ensure that the downloading itself works, we could set up a quick http server in the tests to download a coupler RSRs from just for tha sake of testing.

PR #281, opened 2026-03-30
Prepared for latest RSR data files - with new modis platform names

### LINKED ISSUE #252: MODIS platform name not compatible with satpy
Satpy and the MODIS L1B and L2 files use the platform names "Aqua" and "Terra", but pyspectral uses the old "EOS-Aqua" and "EOS-Terra" names. This results in not finding RSR files (and others?) in pyspectral.

adybbroe:
Platform names are Aqua and Terra, and *not* EOS-Aqua and EOS-Terra.

 - [x] Closes #252  
 - [ ] Tests added 
 - [ ] Fully documented 


djhoese (approved):
It is possible to have aliases in the various dictionaries in pyspectral so things still work? Or was that already the case? If it was already the case, do the dictionaries need to be updated to point from the new-old name to the old-new name? Otherwise looks good to me.

adybbroe:
> It is possible to have aliases in the various dictionaries in pyspectral so things still work? Or was that already the case? If it was already the case, do the dictionaries need to be updated to point from the new-old name to the old-new name? Otherwise looks good to me.

I guess you could do that quite easily, but we don't have a platform name translation dictionary. I updated the `INSTRUMENT` dictionary to include `Aqua` and `Terra` to be consistent but that doesn't have an effect on this. The rsr filename is created from the `sensor` and `platform_name` as provided. As this is not supposed to happen often I didn't spend more time here. Updated the RSR tar file once more, now also with sensor name in the attributes. And it seems I actually missed including the 2 new RSR files in the new zenodo release. This is fixed now.

PR #294, opened 2026-08-28
fix(seviri): include the first IR row when converting RSR data

### LINKED ISSUE #253: Missing first line of SEVIRI RSR
Hello,

The seviri RSR are extracted at https://github.com/pytroll/pyspectral/blob/main/rsr_convert_scripts/seviri_rsr.py#L115

The call to `sheet.col_values(1, start_rowx=13, end_rowx=113))` is 0-indexed (look at https://github.com/python-excel/xlrd/blob/master/xlrd/sheet.py#L567 ). -> Line 13 in the spreadsheet view is missing in the data file.

For SEVIRI on Meteosat-11, band `IR7.3`, the first wavelength is 6.37 whereas in the spreadsheet it is 6.35.

Could the SEVIRI parsing code be modified to use one extra line?

I did a test for this case and the conversion program happily adds this first line. If this is "accepted", I can make a PR of course.

Manny7717:
## What

Fixes #253: the SEVIRI RSR converter skips the first IR spreadsheet data row. The issue report was correct: `xlrd` is 0-indexed, so spreadsheet line 13 corresponds to `start_rowx=12`, not `13`.

For `IR7.3` in the bundled workbook, row 12 starts at wavelength `6.35`, while the current loader starts at row 13 and incorrectly begins at `6.37`.

## Change

In `rsr_convert_scripts/seviri_rsr.py`, the IR-channel branch now reads rows `12..111` instead of `13..112` for the wavelength column and every Meteosat response column.

## Verification

- Workbook proof from `pyspectral/data/MSG_SEVIRI_Spectral_Response_Characterisation.XLS`:
  - row 12 col 0 = `6.35`
  - row 13 col 0 = `6.37`
- New regression test `test_ir_channels_include_the_first_spreadsheet_row`
  - **fails before the fix**: loaded first wavelength is `6.37`
  - passes after the fix
- Test runs:
  - `.venv/bin/python -m pytest pyspectral/tests/test_seviri_rsr.py pyspectral/tests/test_config.py -q` -> **4 passed**
  - `.venv/bin/flake8 rsr_convert_scripts/seviri_rsr.py pyspectral/tests/test_seviri_rsr.py pyspectral/tests/test_config.py` -> **passed**

## Why this is scoped correctly

This change only touches the IR-specific row window in the shared loader path. The non-IR channels already use `start_rowx=12`, and HRV uses a different block entirely.

### DIFF
--- a/pyspectral/tests/test_seviri_rsr.py
+++ b/pyspectral/tests/test_seviri_rsr.py
@@ -1,0 +1,41 @@
+import importlib
+import sys
+import types
+from pathlib import Path
+
+from xlrd import open_workbook
+
+
+def _import_seviri_module(repo_root: Path):
+    sys.modules.pop("rsr_convert_scripts.seviri_rsr", None)
+    sys.modules["pkg_resources"] = types.SimpleNamespace(
+        resource_filename=lambda package, resource: str(repo_root / "pyspectral" / "data")
+    )
+    return importlib.import_module("rsr_convert_scripts.seviri_rsr")
+
+
+def test_ir_channels_include_the_first_spreadsheet_row(monkeypatch):
+    repo_root = Path(__file__).resolve().parents[2]
+    xls_path = repo_root / "pyspectral" / "data" / "MSG_SEVIRI_Spectral_Response_Characterisation.XLS"
+    workbook = open_workbook(str(xls_path))
+    sheet = workbook.sheet_by_name("IR7.3")
+
+    # The issue's concrete repro: the first wavelength/data row is line 13 in the
+    # spreadsheet UI, which is row index 12 for xlrd.
+    assert sheet.cell_value(12, 0) == 6.35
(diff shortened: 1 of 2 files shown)


Manny7717:
pre-commit.ci was right — my new test file was missing module/function docstrings (D100/D103 under flake8-docstrings). Added them in 7eb0722 and verified locally with the exact hook environment (flake8 + docstrings + bugbear + debugger): clean, and the regression test still passes.

djhoese (commented):
@adybbroe will need to be the final say on this pull request, especially since it will mean regenerating the RSR data and uploading it to the LUT data host. I think this looks good from what I understand except that this would be the first RSR script with tests. These aren't really user facing and are run "once" so we don't typically have tests for them. I almost want to say the test module should be removed, but @adybbroe should make the final call on this too. Thanks for fixing this @Manny7717.

PR #295, opened 2026-08-31
Fix conversion script for v10 of METimage SRF

ninahakansson:
Fix conversion script for v10 of METimage SRF, EPS-SGA1-METIMAGE_SRF_V10.nc

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Fully documented 

### DIFF
--- a/rsr_convert_scripts/metimage_rsr.py
+++ b/rsr_convert_scripts/metimage_rsr.py
@@ -32,11 +32,11 @@ def get_groupname_from_bandname(band):
     groupname = "LVWIR"
     wv = int(band.split("_")[-1])
     if wv > 6000:
-        groupname = "LVWIR"
+        groupname = "Instrument/ISRF/LVWIR"
     elif wv > 1000:
-        groupname = "SMWIR"
+        groupname = "Instrument/ISRF/SMWIR"
     else:
-        groupname = "VISNIR"
+        groupname = "Instrument/ISRF/VISNIR"
     return groupname
 
 


djhoese:
Are the values in the SRFs the same and this is just a group name change?

Manny7717 (commented):
Quick verification notes on head 0c26413:

**Verified locally:** band-name parsing is sound - the script's band keys (e.g. `vii_443`, `vii_865`) parse via `int(band.split("_")[-1])` to wavelengths in nm, and the three thresholds (wv>6000 -> LVWIR, >1000 -> SMWIR, else VISNIR) map every METimage band to the right family (probed all current BANDNAMES entries). The `Instrument/ISRF/<family>` paths are consistent with the v10 file's nested-group layout.

**One compatibility question:** the group-name change is unconditional, but the module docstring (lines 1-9) and rsr_convert_scripts/README.rst:201 still reference the plain `EPS-SGA1_METIMAGE_SRF.nc` (non-v10) file. If any users are still converting the older file (flat `LVWIR`/`SMWIR`/`VISNIR` groups), this change would break their conversion. Options: bump the docstring/README to point at the v10 file explicitly, or make the lookup tolerant (try nested path first, fall back to flat). Could not fetch the EUMETSAT file to double-check the exact v10 group names (anonymous sftp 404), so flagging rather than assuming.

Non-blocking - the fix itself looks correct for v10.

adybbroe:
@djhoese SRFs have been improved. Especially the IR channels differ. Cannot really see any changes in the VIS/NIR bands. I faked @ninahakansson 's new SRFs as Metop-SG-A4 for the convenience of plotting them via:
`python bin/composite_rsr_plot.py --platform_name Metop-SG-A1 Metop-SG-A4 --sensor metimage -r 8 14.0`
rsr_band_0800_1400

adybbroe (approved):
Thanks for updating the SRFs to the latest from EUMETSAT!
LGTM

